PR #77, opened 2014-06-05
Rewrite json converter test

fer-rum:
I tried to avoid messing with files others might need. Still double check that I didn't mess up something.


mkroetzsch:
> This is stuck in limbo for half a year now. A lot has happened since then. Do we still care?

Good question. I suppose the JSON format (and our code for processing it) has changed so much that the modifications do not make sense any more. Our Jackson code is relatively well tested. More tests would mainly be needed for the JsonDumpFileProcessor (under wdtk-dumpfiles).

Shall we close this pull request unmerged, or is there anything we could try to convert to the current code?

fer-rum:
The expiration date of this pull request was long ago, so I am closing it.

PR #87, opened 2014-07-06
Conversation comand line client

guenthermi:
This branch include a new module wdtk-clt with a command line client to serialize json and several rdf dumps from the data of wikibase dumps


wdtk-distribution/pom.xml
```diff
@@ -34,14 +41,25 @@
...
+				<configuration>
+					<descriptorRefs>
+						<descriptorRef>jar-with-dependencies</descriptorRef>
+					</descriptorRefs>
+					<appendAssemblyId>false</appendAssemblyId>
+				</configuration>
 				<executions>
 					<execution>
-						<id>create-binary-jar</id>
+						<id>make-assembly</id>
```
mkroetzsch: What's the reason for this change? Could this break anything that worked before?

PR #96, opened 2014-09-09
Serialization of constraints, package 'model'

julianmendez:
This branch is about serialization of constraints. It contains package 'model' and its unit test classes.


julianmendez:
The pull request on this branch has been suspended because it needs an update before being merged into master.

PR #98, opened 2014-10-09
Create parser of constraints

mkroetzsch:


julianmendez:
This branch is ready to be reviewed.

julianmendez:
This branch uses flexible bit vectors now. This is useful when the maximum needed size of a bit vector is unknown at moment of its creation.

ricordisamoa:
How does this relate to the recent work on https://github.com/wikimedia/mediawiki-extensions-WikidataQualityConstraints?

mkroetzsch:
@ricordisamoa Our code was only about extracting template-based constraint information to make it accessible in Java. The WikidataQualityConstraints extension is focussed on checking the constraints. THe main reason why our work on this has stalled is that it is planned (for some time now) to move (some) constraints from templates to statements on property pages, which makes a template-parsing approach obsolete. However, right now it is not clear what the status of this activity is. Maybe WikidataQualityConstraints also does some work in this direction.

JeroenDeDauw:
250+ commits and open since 2014? wtf?

mkroetzsch:
With the conversion of constraint templates into Wikidata properties, this code is no longer needed. The pull request can therefore be closed.

PR #110, opened 2014-12-29
Serializable datamodel pull

fer-rum:
This pull request contains:

* Serializable implementation for the standard datamodel
* Extension to the ```DatamodelConverter``` that allow true deep cpoies (Draft)
* A ```TypeInspector``` class that simply can list the types involved in the structure of a given ```ItemDocument```. This was used for debugging purposes, but I found it a nice idea, so I included the draft
* A fix for the unknown JsonProperty in the ```JacksonPropertyDocument``` (again? I think it should already be in master)


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/TypeInspector.java
```diff
@@ -0,0 +1,110 @@
...
+import org.wikidata.wdtk.datamodel.interfaces.SnakVisitor;
+import org.wikidata.wdtk.datamodel.interfaces.SomeValueSnak;
+import org.wikidata.wdtk.datamodel.interfaces.Statement;
+import org.wikidata.wdtk.datamodel.interfaces.StatementGroup;
+import org.wikidata.wdtk.datamodel.interfaces.StringValue;
+import org.wikidata.wdtk.datamodel.interfaces.TimeValue;
+import org.wikidata.wdtk.datamodel.interfaces.ValueSnak;
+import org.wikidata.wdtk.datamodel.interfaces.ValueVisitor;
+
+// TODO this is still completely rudimentary
```
mkroetzsch: TODO: document

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/TypeInspector.java
```diff
@@ -0,0 +1,110 @@
...
+public class TypeInspector implements SnakVisitor<String>, ValueVisitor<String> {
+
+	public String inspect(ItemDocument document) {
+		String result = document.getClass().getName();
```
mkroetzsch: Probably use a StringBuilder

mkroetzsch:
Too many things in one pull request here ...

I think deepcopy should be the default for our copy behaviour. One can add configuration options to the DatamodelConverter object for changing this. I will make this change in this branch now.

I don't think the TypeInspector is ready for merge. At least it should have some documentation on what it is supposed to do. A test that tests that a complex object can be converted to a string by this class would also make sense (not just to bring up test coverage a bit).

The significant decrease in test coverage is probably due to the fact that none of the new code is tested. Do you have any idea on how serializability is usually tested? There should be an easy way.

mkroetzsch:
As part of the changes I want to do, I also merge with the new master now, which seems to be some work.

mkroetzsch:
> Do you have any idea on how serializability is usually tested?

I retract this question. Serializable does not include any code that requires testing.

Issue #106, opened 2014-12-11
Properties now have statements 


shilad:
See https://www.wikidata.org/wiki/Wikidata:Project_chat#statements_on_properties_are_live_.28.2Bdeletion_propagation.29

mkroetzsch:
Fixed in master by merging #111.

PR #111, opened 2014-12-29
Issue #106: support property statements

mkroetzsch:
Support property statements; more robust JSON parsing

* Statements on property documents now supported in datamodel
* Parsing property statements from JSON supported
* more robust JSON dump parsing: recover from broken entities

This addresses issue #106.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,6 +4,10 @@ Wikidata Toolkit Release Notes
 Version 0.4.0
 -------------
 
+New features:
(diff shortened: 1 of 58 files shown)


shilad:
Hooray!

fer-rum:
Ignore my comment about the legacy Json. After some time pondering about the case, I found that it should indeed be fine since it simply (but silently) ignores what it doesn't know about. And anyway which madman would encode the newer dumps with their new features in the legacy format?

PR #117, opened 2015-01-21
Use Jackson for all JSON work -- remove JSON.org dependency

mkroetzsch:
* JsonSerializer in datamodel.json.jackson replaces the old
JsonSerializer in datamodel.json
* All code for parsing/serializing old JSON goes away
* RDF property type lookup now done via Jackson too
* POM files updated to remove org.json dependency


mkroetzsch:
@guenthermi This code removes datamodel.json.JsonSerializer but it introduces datamodel.json.jackson.JsonSerializer instead (mostly same interface). It should be easy to change this in the command line client as well.

mkroetzsch:
As expected, deleting well-tested code decreases coverage ... the new code should be well tested though.

PR #119, opened 2015-02-12
Track the progress of Web downloads

earldouglas:
I noticed that some downloads (e.g. the full JSON dump) take a while, leaving me wondering about the download progress.

This adds the ability to track the progress of Web-based InputStream instances as they are downloaded.  Progress is logged to a Logger by percent completion.

### DIFF
--- a/wdtk-util/src/main/java/org/wikidata/wdtk/util/ProgressLoggingInputStream.java
+++ b/wdtk-util/src/main/java/org/wikidata/wdtk/util/ProgressLoggingInputStream.java
@@ -1,0 +1,81 @@
+package org.wikidata.wdtk.util;
+
+/*
+ * #%L
+ * Wikidata Toolkit utilities
+ * %%
+ * Copyright (C) 2014 Wikidata Toolkit Developers
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


earldouglas:
Progress looks like this:

```
2015-02-12 14:07:48 INFO  - Downloading JSON dump file 20150209.json.gz from http://dumps.wikimedia.org/other/wikidata/20150209.json.gz ...
2015-02-12 14:10:22 INFO  - Retrieved 1% of http://dumps.wikimedia.org/other/wikidata/20150209.json.gz
2015-02-12 14:12:59 INFO  - Retrieved 2% of http://dumps.wikimedia.org/other/wikidata/20150209.json.gz
```

An alternate implementation might use a separate thread to log every X seconds.  This would be better for fast downloads, since progress tracking is less useful and noise should be avoided, and would be better for slow downloads, since it might be minutes between each percent increase, still leaving the user to wonder.

mkroetzsch:
Thanks, this is very useful. We will try to merge this asap. I also just pushed a small change that ensures that partial downloads of JSON dumps use a temporary file name, so that aborted downloads do not lead to corrupted files being stored locally.

mkroetzsch:
I had a similar feature implemented in an early prototype, but my strategy there was to report progress on another level: rather than directly in the input stream, I was doing it when reading from that stream into a file. At this place, it is possible to read data in chunks of arbitrary size. Right now, we use the maximal long as a size, so we read everything in one chunk, but one could also read in chunks of 1%. An advantage would be that one does not need to execute code after every byte that is fetched, but I also think that if would require less code overall. Another idea could be to read in chunks of 10MB and to check after each such chunk if another percent was achieved *or* if a certain time has passed, and to log some output in either case. This would be a simple way to achieve time-based reports without the need for a timer thread. And we already have a Timer implementation that one could use.

Another thing to do is related to testability (this applies in any cases as long as we want percent progress to be printed [which is very nice]). The only problem there is the code of getContentLength(), which cannot be tested in a controlled way (one would need to access online data). A good way to make this method mockable would be to add it to our existing WebResourceFetcher object that is now used to encapsulate Web access (and for which there is also a mock implementation).

earldouglas:
This needs to be rebased. Any interest in it, or shall I close it?

PR #126, opened 2015-03-11
Activity report

guenthermi:
This Branch add functions to let the client print a report file which include the names / locations of the generated files (with statistical information about them) after all actions got finished.


mkroetzsch:
Oops, this request cannot be merged automatically right now.

wdtk-client/src/test/java/org/wikidata/wdtk/client/ClientTest.java
```diff
@@ -83,6 +83,23 @@ public void testQuietLoggingConfig() throws ParseException, IOException {
 	}
 
 	@Test
+	public void testJsonOutput() {
+		DumpProcessingAction action = new JsonSerializationAction();
+		action.open();
+		action.close();
+		assertEquals(action.getReport(),
+				"Finished serialization of 0 EntityDocuments in file {PROJECT}-{DATE}.json");
```
mkroetzsch: Why does the report contain a file-name pattern instead of a real file name?

wdtk-client/src/main/java/org/wikidata/wdtk/client/RdfSerializationAction.java
```diff
@@ -251,4 +254,21 @@ private void printHelp() {
 					+ TASK_HELP.get(supportedTask), 75, "\n   ", true));
 		}
 	}
+
+	@Override
+	public String getReport() {
+		String message = "Finished serialization of "
+				+ this.serializer.getTripleCount() + " RDF triples in file "
+				+ this.outputDestination;
```
mkroetzsch: Related to my comment on the JSON test: this should not use this.outputDestination but the same file name as is really used. This is computed in the method createRdfSerializer above. This computation should be moved to another function so that it can be fetched from several places (and this function should call insertDumpInformation to replace the placeholders).

wdtk-client/src/main/java/org/wikidata/wdtk/client/JsonSerializationAction.java
```diff
@@ -97,4 +103,21 @@ public void close() {
...
+		String message = "Finished serialization of "
+				+ this.serializer.getEntityDocumentCount()
+				+ " EntityDocuments in file " + this.outputDestination;
+		return message;
+	}
+
+	@Override
+	public String getActionName() {
+		if (this.name != null) {
+			return name;
```
mkroetzsch: should be: this.name

wdtk-client/src/main/java/org/wikidata/wdtk/client/DumpProcessingAction.java
```diff
@@ -101,4 +101,26 @@
...
+
+	/**
+	 * Sets the actions name.
+	 * 
+	 * @param name
+	 */
+	void setActionName(String name);
+
+	/**
+	 * Returns a name which terms the action.
```
mkroetzsch: Analogous to the setActionName documentation, this should read "Returns the name of the action.".

wdtk-client/src/main/java/org/wikidata/wdtk/client/DumpProcessingAction.java
```diff
@@ -101,4 +101,26 @@
...
+	/**
+	 * Returns a report message containing information about the files which
+	 * were generated.
+	 * 
+	 * @return report message
+	 */
+	String getReport();
+
+	/**
+	 * Sets the actions name.
```
mkroetzsch: Would need to be "the action's name" or "the name of the action". I would also add "If this is not set, a default name will be used."

wdtk-client/src/main/java/org/wikidata/wdtk/client/DumpProcessingAction.java
```diff
@@ -29,16 +29,16 @@
...
- *
+ * 
  */
 public interface DumpProcessingAction extends EntityDocumentDumpProcessor {
 
 	/**
 	 * Returns true if processing requires a {@link Sites} object to be set.
 	 * This is the case for some operations that process site links.
-	 *
+	 *
```
mkroetzsch: Any idea why these changes are happening? I am also seeing this on my machine and it creates a lot of pointless changes. Is Maven doing this or Eclipse?

wdtk-client/src/main/java/org/wikidata/wdtk/client/RdfSerializationAction.java
```diff
@@ -251,4 +254,21 @@ private void printHelp() {
...
+		return message;
+	}
+
+	@Override
+	public String getActionName() {
```
mkroetzsch: Could we move this to the abstract super class and use the local name of this.class() as a default? Alternatively, could we just require a default action name as a constructor parameter to the abstract super class. Then the superclass can implement getActionName() (once and for all) without having to repeat this code in each subclass. But using the local class name seems even simpler.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -151,6 +155,12 @@ public void performActions() {
 		openActions();
 		this.dumpProcessingController.processDump(dumpFile);
 		closeActions();
+		try {
+			createReport();
+		} catch (IOException e) {
+			logger.error("Could not print report file!");
```
mkroetzsch: Please add the exception message to the logged string.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +214,32 @@ private void openActions() {
 	}
 
 	/**
+	 * Creates a report file including the results which came from the
+	 * {@link DumpProcessingAction#getReport()} methods or printing it to
+	 * stdout.
+	 * 
+	 * @throws IOException
+	 */
+	void createReport() throws IOException {
```
mkroetzsch: Should be called "writeReport". Functions that start with "create..." should return a (newly created) object.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +214,32 @@ private void openActions() {
 	}
 
 	/**
+	 * Creates a report file including the results which came from the
+	 * {@link DumpProcessingAction#getReport()} methods or printing it to
+	 * stdout.
```
mkroetzsch: Please review this sentence (word order is mixed up). Also, it seems that you are never printing to stdout; you use the logger.

wdtk-client/src/main/java/org/wikidata/wdtk/client/ClientConfiguration.java
```diff
@@ -252,18 +267,22 @@ public boolean getOfflineMode() {
...
 	 * to stdout.
-	 *
+	 * 
 	 * @return true if the application should not log messages to stdout
 	 */
 	public boolean isQuiet() {
 		return this.quiet;
 	}
 
+	public String getReportFilename() {
```
mkroetzsch: Documentation missing.

wdtk-client/src/main/java/org/wikidata/wdtk/client/ClientConfiguration.java
```diff
@@ -95,6 +95,10 @@
 	 * Short command-line alternative to {@link #OPTION_OUTPUT_STDOUT}.
 	 */
 	public static final String CMD_OPTION_OUTPUT_STDOUT = "s";
+	/**
+	 * Short command-line alternative to {@link #OPTION_OUTPUT_REPORT}.
```
mkroetzsch: Wrong documentation

wdtk-client/src/main/java/org/wikidata/wdtk/client/ClientConfiguration.java
```diff
@@ -142,6 +146,11 @@
 	 */
 	public static final String OPTION_FILTER_PROPERTIES = "fProp";
 	/**
+	 * Name of the option to print a report about the files produced by
+	 * DumpProcessingOutputActions
```
mkroetzsch: Misleading documentation. The option is not "to print a report" but to define the output file to print the report to. The report will only be "printed" if the option is not set, right?

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +214,32 @@ private void openActions() {
...
+	 */
+	void createReport() throws IOException {
+		StringBuilder builder = new StringBuilder();
+		for (DumpProcessingAction action : this.clientConfiguration
+				.getActions()) {
+			builder.append(action.getActionName() + ": ");
+			builder.append(action.getReport());
+			builder.append(System.getProperty("line.separator"));
+		}
+		if (this.clientConfiguration.reportFilename != null) {
```
mkroetzsch: Use the getter; don't read package private fields directly.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +214,32 @@ private void openActions() {
...
+			File reportFile = new File(this.clientConfiguration.reportFilename);
```
mkroetzsch: use the getter

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +214,32 @@ private void openActions() {
...
+			BufferedWriter writer = new BufferedWriter(new FileWriter(
+					reportFile));
+			writer.write(builder.toString());
+			writer.close();
+		} else {
+			logger.info((builder.toString()));
```
mkroetzsch: Why are there double parentheses?

mkroetzsch: Wouldn't it be nicer to have one info log message per action instead of having each of them in one report?

wdtk-client/src/main/java/org/wikidata/wdtk/client/DumpProcessingOutputAction.java
```diff
@@ -93,9 +93,11 @@
 	 */
 	protected Set<Closeable> outputStreams = new HashSet<>();
 
+	protected String name;
```
mkroetzsch: Missing documentation.

wdtk-client/src/main/java/org/wikidata/wdtk/client/RdfSerializationAction.java
```diff
@@ -251,4 +254,21 @@ private void printHelp() {
...
+		if (this.name != null) {
+			return name;
```
mkroetzsch: this.name

wdtk-client/src/test/java/org/wikidata/wdtk/client/ClientTest.java
```diff
@@ -83,6 +83,23 @@ public void testQuietLoggingConfig() throws ParseException, IOException {
...
+	}
+
+	@Test
+	public void testRdfOutput() {
+		DumpProcessingAction action = new RdfSerializationAction();
+		action.open();
+		action.close();
+		assertEquals(action.getReport(),
+				"Finished serialization of 24 RDF triples in file null");
```
mkroetzsch: This is not a robust test. The number 24 will change as soon as we change the vocabulary.

wdtk-client/src/test/java/org/wikidata/wdtk/client/ClientConfigurationTest.java
```diff
@@ -85,6 +85,7 @@ public void testDefaultArguments() {
 		assertEquals(config.getFilterLanguages(), null);
 		assertEquals(config.getFilterSiteKeys(), null);
 		assertEquals(config.getFilterProperties(), null);
+		assertEquals(config.getReportFilename(), null);
```
mkroetzsch: General comment: assertEquals should always have the expected value in the first argument, and the tested value in the second. This is wrong in much of our existing code, but we should try to fix it when we touch test methods or create new ones.

wdtk-client/src/test/java/org/wikidata/wdtk/client/JsonSerializationActionTest.java
```diff
@@ -60,6 +60,7 @@ public void testDefaults() {
 		assertTrue(action instanceof JsonSerializationAction);
 		assertFalse(action.needsSites());
 		assertTrue(action.isReady());
+		assertEquals(action.getActionName(), "JsonSerializationAction");
 	}
```
mkroetzsch: Currently we are missing tests that test the actual report output. You are testing the report string, but not its writing to file. This could be added below by setting a report file for one of the tests, and making sure that it contains a line starting with the action name. It would also be useful to set a non-standard action name to test this in action as well. If you do this in only one of the methods that test JSON output, both report file and logger report will be tested.

wdtk-client/src/test/java/org/wikidata/wdtk/client/RdfSerializationActionTest.java
```diff
@@ -38,6 +38,7 @@ public void testDefaults() {
 		assertTrue(action instanceof RdfSerializationAction);
 		assertTrue(action.needsSites());
 		assertTrue(action.isReady());
+		assertEquals(action.getActionName(), "RdfSerializationAction");
 	}
```
mkroetzsch: Same comment as for JSON, but probably it's enough to test report file writing once since the code is shared.

mkroetzsch:
@guenthermi I have added a number of inline comments. Once these are fixed, the branch should be ready to merge.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
guenthermi: In the prepareSites method the exception message is also missing. Should I change this there too?

wdtk-client/src/test/java/org/wikidata/wdtk/client/JsonSerializationActionTest.java
guenthermi: I don't understand what you mean with "setting a report file". The report file is created by the createReport method in the Client. The report messages consisting of the action name and the getReport result are generated in this method, too. So I don't know to test it. It is also not possible to mock this, or something like this...

wdtk-client/src/main/java/org/wikidata/wdtk/client/RdfSerializationAction.java
guenthermi: We could also use a Getter for an default action name. So we can move the method to the abstract super class and the default action name need not depend on the class name.

wdtk-client/src/test/java/org/wikidata/wdtk/client/JsonSerializationActionTest.java
mkroetzsch: I see, so it's another file to test. But why can't you test the method in the Client? If your code is really untestable, then it should be changed to be testable (at least I don't see any function here that cannot be tested in principle). Indeed, with the code in the client, you could even create a mock action that creates a predefined report so that you can test the exact string. But I agree that this test should be part of the client test file, not of this action's test file.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
mkroetzsch: Yes, please do.

wdtk-client/src/main/java/org/wikidata/wdtk/client/RdfSerializationAction.java
mkroetzsch: Yes, that would be an acceptable solution too. I think, however, that such additional get methods are not necessary here: I don't see why one should return something else than the class name. It seems like a good default, and subclasses could still overwrite it by calling the setter in their constructor with another string.

wdtk-client/src/main/java/org/wikidata/wdtk/client/DumpProcessingAction.java
guenthermi: I think this is caused by the eclipse auto-formatter. Maven do not make such changes.

mkroetzsch: Could you try to switch this off in your Eclipse? When I safe a file, it removes these spaces. The formatter (SHIFT+CTRL+F) does not change the spaces in my installation. As it is now, we are getting a lot of distracting changes in the diffs based on this.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +215,35 @@ private void openActions() {
...
+				.getActions()) {
+			if (this.clientConfiguration.getReportFilename() != null) {
+				builder.append(action.getActionName() + ": ");
+				builder.append(action.getReport());
+				builder.append(System.getProperty("line.separator"));
+			} else {
+				logger.info(action.getActionName() + ": " + action.getReport());
+			}
+		}
+		if (this.clientConfiguration.getReportFilename() != null) {
```
mkroetzsch: Move up to the if we already have.

wdtk-client/src/main/java/org/wikidata/wdtk/client/Client.java
```diff
@@ -204,6 +220,45 @@ private void openActions() {
...
+				outputDirectory = Paths.get(".");
+			}
+			DirectoryManager dm = DirectoryManagerFactory
+					.createDirectoryManager(outputDirectory);
+			OutputStream out = dm.getOutputStreamForFile(Paths
+					.get(this.clientConfiguration.getReportFilename())
+					.getFileName().toString());
+
+			OutputStream bufferedFileOutputStream = new BufferedOutputStream(
+					out, 1024 * 1024 * 5);
```
mkroetzsch: That's a big buffer for the three lines of report you normally have. Maybe just use the default.

Issue #125, opened 2015-03-05
Extend bit vector to be more flexible with respect to bounds


julianmendez:
The current default implementation of BitVector (BitVectorImpl) throws and exception when trying to access a position out of the original bounds. A new configuration, new method, or new class implementation could let the bit vector grow as needed. This is especially helpful when processing dumps.

PR #127, opened 2015-03-12
Make bit vector more flexible

julianmendez:
The purpose of these changes is to have a more flexible implementation of bit vector. This means that elements in non-negative positions outside the bit vector are considered 'false', and a bit can be directly stored at any non-negative position without the need of resizing the vector ( issue #125 ).


wdtk-storage/src/main/java/org/wikidata/wdtk/storage/datastructures/BitVectorImpl.java
```diff
@@ -162,16 +170,32 @@ public boolean addBit(boolean bit) {
...
+	 * given position. If the bit vector needs to be enlarged, new
+	 * <code>false</code> elements are added.
+	 * 
+	 * @param position
+	 *            position
+	 */
+	void ensureSize(long position) {
+		assertNonNegativePosition(position);
+		while (position >= this.size) {
+			addBit(false);
```
mkroetzsch: Isn't this very inefficient? It seems that this loop might easily be executed 100 million times in a typical run. Can't we just increase the size in one step?

PR #134, opened 2015-03-30
Add rank

guenthermi:
This adds functions to export the statement ranks to the rdf module.

### DIFF
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
@@ -364,18 +364,21 @@ public class RdfConverter {
 		}
 	}
 
+	/**
+	 * Writes a triple for the {@link StatementRank} of a {@link Statement} to
+	 * the dump.
+	 * 
+	 * @param subject
+	 * @param rank
+	 */
 	void writeStatementRankTriple(Resource subject, StatementRank rank) {
 		String rankRepresentation = rankToRdfRepresentation(rank);
-		if (rankRepresentation != null){
-			try {
-				this.rdfWriter.writeTripleStringObject(subject, RdfWriter.WB_RANK,
-						rankRepresentation);
-			} catch (RDFHandlerException e) {
-				logger.warn("Could not write rank " + rank.toString()
-						+ " of statement " + subject.toString());
-			}
-		}else{
-			logger.warn("Unkown rank " + rank.toString());
+		try {
+			this.rdfWriter.writeTripleUriObject(subject, RdfWriter.WB_RANK,
+					rankRepresentation);
+		} catch (RDFHandlerException e) {
+			logger.warn("Could not write rank " + rank.toString()
+					+ " of statement " + subject.toString());
 		}
 	}
 
@@ -490,16 +493,22 @@ public class RdfConverter {
 		return ((this.tasks & task) == task);
 	}
 
+	/**
+	 * Creates an URI which represents the statement rank in a triple.
+	 * 
+	 * @param rank
+	 * @return
+	 */
 	String rankToRdfRepresentation(StatementRank rank) {
 		switch (rank) {
 		case NORMAL:
-			return "Normal";
+			return Vocabulary.WB_NORMAL_RANK;
 		case PREFERRED:
-			return "Preferred";
+			return Vocabulary.WB_PREFERRED_RANK;
 		case DEPRECATED:
-			return "Deprecated";
+			return Vocabulary.WB_DEPRECATED_RANK;
 		default:
-			return null;
+			throw new IllegalArgumentException();
 		}
 	}
 
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
@@ -365,12 +365,17 @@ public class RdfConverter {
 	}
 
 	void writeStatementRankTriple(Resource subject, StatementRank rank) {
-		try {
-			this.rdfWriter.writeTripleStringObject(subject, RdfWriter.WB_RANK,
-					rank.toString());
-		} catch (RDFHandlerException e) {
-			logger.warn("Could not write rank " + rank.toString()
-					+ " of statement " + subject.toString());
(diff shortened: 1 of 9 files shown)


wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -363,6 +364,21 @@ void writeTermTriples(Resource subject, URI predicate,
...
+		String rankRepresentation = rankToRdfRepresentation(rank);
+		if (rankRepresentation != null){
+			try {
+				this.rdfWriter.writeTripleStringObject(subject, RdfWriter.WB_RANK,
+						rankRepresentation);
+			} catch (RDFHandlerException e) {
+				logger.warn("Could not write rank " + rank.toString()
+						+ " of statement " + subject.toString());
+			}
+		}else{
```
mkroetzsch: strange formatting (no spaces)

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -363,6 +364,21 @@ void writeTermTriples(Resource subject, URI predicate,
 		}
 	}
 
+	void writeStatementRankTriple(Resource subject, StatementRank rank) {
+		String rankRepresentation = rankToRdfRepresentation(rank);
+		if (rankRepresentation != null){
+			try {
+				this.rdfWriter.writeTripleStringObject(subject, RdfWriter.WB_RANK,
+						rankRepresentation);
```
mkroetzsch: Ranks should be represented by resources (URIs) rather than by strings. The URIs should use local names PreferredRank, NormalRank, DeprecatedRank and be defined in the vocabulary class.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -474,4 +490,17 @@ boolean hasTask(int task) {
...
+	String rankToRdfRepresentation(StatementRank rank) {
+		switch (rank) {
+		case NORMAL:
+			return "Normal";
+		case PREFERRED:
+			return "Preferred";
+		case DEPRECATED:
+			return "Deprecated";
+		default:
+			return null;
```
mkroetzsch: Maybe throw an IllegalArgumentException. I don't think the outer code can recover from this anyway.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -363,6 +364,21 @@ void writeTermTriples(Resource subject, URI predicate,
 		}
 	}
 
+	void writeStatementRankTriple(Resource subject, StatementRank rank) {
```
mkroetzsch: Documentation missing.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -474,4 +490,17 @@ boolean hasTask(int task) {
 		return ((this.tasks & task) == task);
 	}
 
+	String rankToRdfRepresentation(StatementRank rank) {
```
mkroetzsch: Documentation missing

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/Vocabulary.java
```diff
@@ -196,6 +196,14 @@
 	}
 
 	/**
+	 * Class for Wikibase rank.
```
mkroetzsch: This is not a class but a property. Properties are defined further down in the file.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -363,6 +364,24 @@ void writeTermTriples(Resource subject, URI predicate,
...
+	 * @param rank
+	 */
+	void writeStatementRankTriple(Resource subject, StatementRank rank) {
+		String rankRepresentation = rankToRdfRepresentation(rank);
+		try {
+			this.rdfWriter.writeTripleUriObject(subject, RdfWriter.WB_RANK,
+					rankRepresentation);
+		} catch (RDFHandlerException e) {
+			logger.warn("Could not write rank " + rank.toString()
+					+ " of statement " + subject.toString());
```
mkroetzsch: Better just throw new RuntimeException(e.toString(), e); as we do in all other cases where we catch this. Should almost never happen and if it does we should better escalate the issue instead of swallowing it silently. Most likely, the file has been closed or the disk is full, so we won't be able to recover from this anyway.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -474,4 +493,23 @@ boolean hasTask(int task) {
 		return ((this.tasks & task) == task);
 	}
 
+	/**
+	 * Creates an URI which represents the statement rank in a triple.
+	 * 
+	 * @param rank
+	 * @return
+	 */
+	String rankToRdfRepresentation(StatementRank rank) {
```
mkroetzsch: Rename to "getUriStringFromRank":
* "Create" should only be used for methods that create new objects where creation has a cost or implies ownership responsibility (note that here you don't even create a String at all)
* "RdfRepresentation" is too vague.

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/RdfConverter.java
```diff
@@ -363,6 +364,24 @@ void writeTermTriples(Resource subject, URI predicate,
...
 
+	/**
+	 * Writes a triple for the {@link StatementRank} of a {@link Statement} to
+	 * the dump.
+	 * 
+	 * @param subject
+	 * @param rank
+	 */
+	void writeStatementRankTriple(Resource subject, StatementRank rank) {
+		String rankRepresentation = rankToRdfRepresentation(rank);
```
mkroetzsch: Rename local variable to match renamed method (see below), or simply inline this call.

PR #137, opened 2015-04-10
Added link outs for ChEMBL and KEGG

egonw:
I could not find unit tests to test if the output via extensions works well, and pointers appreciated so that I can test these extensions too.

### DIFF
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/extensions/ExportExtensions.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/extensions/ExportExtensions.java
@@ -174,6 +174,9 @@ public class ExportExtensions implements ValueVisitor<Void> {
 				SimpleIdExportExtension.newNdlExportExtension(),
 				"http://www.wikidata.org/entity/P349");
 		exportExtensions.registerStringValueExportExtension(
+				SimpleIdExportExtension.newChEMBLExportExtension(),
+				"http://www.wikidata.org/entity/P592");
+		exportExtensions.registerStringValueExportExtension(
 				new FreebaseExportExtension(),
 				"http://www.wikidata.org/entity/P646");
 		exportExtensions.registerStringValueExportExtension(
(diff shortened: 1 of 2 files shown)


mkroetzsch:
Thanks, looks good. Are you sure that the KEGG IRI is really an IRI? It seems that it is the URL of the KEGG web page. Does it also return RDF via content negotiation? The ChEMBL IRI works.

For the future, our hardcoded links should be replaced by a system that fetches the correct IRI pattern from Wikidata. The first step for this would be to introduce a "formatter IRI" property that is used like the existing "formatter URL" (P1630). It would be valuable to file a property creation request for this, so we can update our code in due course. We already have live API support in WDTK now so that we can fetch the property data out-of-order during RDF dump creation, which would allow us to get the "formatter IRI" when needed. Once the new property exists, new IRIs can be added without any changes to the code, and we can simply move our existing IRI patterns to Wikidata (Freebase is an exception, since its IRIs require a change to the ID part that cannot be expressed in a pattern).

egonw:
Mmm... not sure where my comment went, but if you check this page:

    http://www.genome.jp/dbget-bin/www_bget?rp:RP04458

There is a download RDF link which uses this IRI pattern for compounds and drugs (not this one, but I have seen such). BTW, I am not really sure this RDF is particularly useful, but at least confirms their use of this IRI pattern.

I am not sure about the content negotiation, but it does not seem so... also, an a compound page there is no "Download RDF" link yet...

egonw:
I just pinged on Twitter, and Kozo Nishida informed me about this:

  http://reactionontology.org/

So, I ran a "describe <>" on the SPARQL end point also confirming the use of the IRI pattern:

  http://www.genome.jp/sparql/reactionontology/?default-graph-uri=http%3A%2F%2Freactionontology.org&query=describe+%3Chttp%3A%2F%2Fwww.kegg.jp%2Fentry%2FC00025%3E&format=text%2Fturtle&timeout=0&debug=on

mkroetzsch:
Ok, fair enough. Then let's just keep it for now and hope that they will provide proper content negotiation in the future. We can always change the IRIs if needed.

mkroetzsch:
@egonw We have now filed a property creation request for a new property that will make it easier to add links to external RDF datasets in the future:

https://www.wikidata.org/wiki/Wikidata:Property_proposal/Property_metadata#formatter_URI

Support and comments are welcome.

PR #140, opened 2015-04-23
Added support for fetching entities by sitelink. Fixes #138. 

### LINKED ISSUE #138: Wikibase API: Support fetching entities by sitelink
The WikibaseDataFetcher class can retrieve entity data for a list of entity ids. The API also supports the retrieval of entities by their article title in Wikipedia (or any other linked site). This should be supported in WDTK as well.

For example, the API can retrieve the entity data for Dresden as follows:

https://www.wikidata.org/w/api.php?action=wbgetentities&format=json&sites=enwiki&titles=Dresden 

This way of querying is not currently implemented in Wikidata Toolkit, but it would be easy enough to add based on the code we already have. The relevant code is in the Java file

https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataFetcher.java

The main method is getEntityDocuments(List<String> entityIds). This code constructs a request URL based on a list of ids and then parses the result. In order to support requests by sitelink, one has to add a new method that takes a sitename and a list of titles, constructs the request URL as above, and then uses the exact same code for reading the result (the result format is the same).

Pull requests are welcome.

alansaid:


mkroetzsch:
Looking good, many thanks.

mkroetzsch:
In fact, I just did a few small changes to the code which may affect your use of it (if you already are using it):
* Methods are renamed (getEntityDocumentsByTitle)
* The resulting HashMap for this method really is indexed by titles as documented (not by Qids as it was)
* Additional convenience methods for getting documents by title
* Extended example program to show how this works

PR #141, opened 2015-04-23
Link properties to property documents. Fixes #84.

### LINKED ISSUE #84: Missing linkage between generated properties
We are trying to use RDF data from WikiDataToolKit, and we noticed the generated properties like those shown below:

original property              generated properties
ns : P01230            ->           ns: P01230c 
ns : P01230            ->           ns: P01230s
ns : P01230            ->           ns: P01230v
ns : P01230            ->           ns: P01230r
ns : P01230            ->           ns: P01230q

We have found out that there were no linkages between original properties and generated properties. Actually we would like to use some information such as labels, descriptions from original properties, so we think there could be an addition
to org.wikidata.wdtk.rdf.OwlDeclarationBuffer.writePropertyDeclarations(), which makes some links such as:

<http://www.wikidata.org/entity/P1230>  ns:simpleClaimProperty <http://www.wikidata.org/entity/P1230c>

Thank you for making the toolkit.

guenthermi:
this adds an task for exporting links between wikidata properties and there corresponding rdf properties according to issue #84


mkroetzsch:
@guenthermi It seems that my first merge of subpropertyof made this one unmergeable :-(. Could you have another look to see if this can be fixed?

guenthermi:
Ok there was a little merge conflict. Now its fixed.

PR #143, opened 2015-04-25
Subpropertyof

guenthermi:
This adds an task to export subpropertyOf statements.


mkroetzsch:
Looking good in general, but I think we should use a single datatype registry that caches known types instead of retrieving them every time we need them from the Web. Could we use the datatype registry that we already have?

PR #148, opened 2015-06-14
Recent changes

mardam:
This solution fetches and parses recent changes from the RSS feed.


mkroetzsch:
This should be updated to use the new ApiConnection before being merged.

wetneb (commented):
I could imagine this being still useful: parsing the recent changes is still the only way one can stay in sync with a Wikibase instance, out of the box.

I would be especially interested in some logic for recent changes polling (generating a real-time stream of events, where each event appears in order, exactly once). This PR does not do this yet, but I imagine the classes it introduces could be a good basis for that.

If there is interest I could try to take over this PR and get it merged. Thoughts? @Tpt @robertvazan

robertvazan:
@wetneb  I will definitely need to read recent changes in my project at some point in the future. I would certainly test this and maybe I would find something to contribute.

Issue #130, opened 2015-03-24
Wikibase API: support wbgetentities for more than 50 entities


mkroetzsch:
By default, the Wikibase API only returns at most 50 entities via wbgetentities (for anonymous users). The WDTK API implementation currently ignores this limit. It should be modified to split bigger lists into several request (merging the results into one result list).

wetneb:
I think this was fixed by b4179bacc81c3bad5aaee97329034332f99a65ea:
https://github.com/Wikidata/Wikidata-Toolkit/blob/283f4f18813c05e906fcfa1f8daa220d50c2c208/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataFetcher.java#L167-L187

PR #154, opened 2015-07-01
Wb get entity

mardam:
This is a solution for issue #130. It considers the maximal size of 50 entities per request of the Wikibase API.


JeroenDeDauw:
Should this branch be deleted as there is a new PR?

Tpt:
I assume that we could delete this branch because the other PR seems to provide the same features.

PR #157, opened 2015-07-28
Wikibaseapi extention

guenthermi:
This provides a Wikibase API interface for WDTK.


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -0,0 +1,510 @@
...
+			this.username = "";
+			this.password = "";
+		}
+	}
+
+	/**
+	 * Reads out the Set-Cookie Header Fields and fills the cookie map of the
+	 * API connection with it.
+	 * <p>
+	 * TODO Why does this method clear all existing cookies?
```
mkroetzsch: I do not understand why the fillCookies method deletes all existing cookies. This seems to be a potential source of errors. I would expect that cookies can always be overwritten by sending a new cookie with a given key.

mkroetzsch:
The code is currently still not very well tested. The problem is the URLConnection, but one can use Mockito to mock it. Splitting methods into smaller ones (e.g. one method to create the URLConnection, and one to populate it with data) would make it possible to test most code with mocked objects. I already changed the setCookies method to return the cookie string without really setting it (the caller can set it). This makes it possible to test this method without any URLConnection (not even a mocked one). This approach might work in other cases too.

mkroetzsch:
P.S. The commenting function of coveralls is somehow broken for our project. See https://coveralls.io/github/Wikidata/Wikidata-Toolkit for coverage reports. As you can see, coverage drops by more than 1% when merging this.

mkroetzsch:
I have merged master to confirm that the test case has been fixed now. The absolute coverage has not changed but is now shown as a decrease because of the improved coverage in master (which is back at 90%, as it should be).

Issue #136, opened 2015-04-01
Make it possible to process local dump files


mkroetzsch:
DumpFileProcessor is the main class for processing dumps and for configuring entity processors and filters. However, right now this class can only process dumps that have been downloaded from the Web (or are put into the download file structure artificially). Even without using DumpFileProcessor, there is no direct way of processing any arbitrary local file: even the MwDumpFileProcessor classes require an MwDumpFile object as input, which is not easy to create for an arbitrary local file. Similarly, the command-line client does not provide any option of specifying a particular local file to process.

Todos:
* A custom local dump file class (implementation of MwDumpFile) should be added. It should be created from a single file name and other metadata required for the interface.
* Such a file could then be used with the DumpProcessingController's processDump() method. An example should be added to illustrate this.
* The command-line client should receive an option to use local dump files.

alansaid:
Is there an ETA on this functionality? I'd be happy to try and implement this myself and contribute it back (although I just started using this toolkit now so not guaranteeing anything)

mkroetzsch:
We have not started working on this yet, so feel free to give it a try along the lines I suggested above. Let me know if you need more hints. In essence, a single, standalone implementation of the MwDumpFile interface is needed, so you hopefully don't have to read much code overall. If you have a pull request, I am happy to review it for merging into master.

mkroetzsch:
This has been fixed in master now. A new example program shows how to process arbitrary local dumps. The command-iine client has a new option "--input" to specify a path to a file.

PR #161, opened 2015-08-04
Local dump file#136

mardam:
This is a solution for issue #136 . It makes it possible to process dumps locally and also contains an example that illustrates how to process a file locally.


mkroetzsch:
The code fails with NullPointerExceptions if prepareDumpFile() is not called explicitly.

PR #168, opened 2015-09-04
Use TravisCI cache for maven

JeroenDeDauw:
See http://docs.travis-ci.com/user/caching/

Not entirely sure the path should be `$HOME/.m2` rather than `$HOME/.m2/repository`. The last comment [here](https://github.com/travis-ci/travis-ci/issues/1441) suggests the former works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,9 @@ jdk:
 
 after_success:
   - mvn clean cobertura:cobertura coveralls:cobertura
+
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.m2


mkroetzsch:
The page on travis says that "The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure" and that this is "still experimental" anyway.

Also, what is this supposed to do? Would this speed up our builds on travis?

JeroenDeDauw:
The new container based infrastructure is used when TravisCI detects it can; ie when there is no sudo usage. I added the explicit config to turn sudo off, so the new stuff will certainly be used.

> Also, what is this supposed to do? Would this speed up our builds on travis?

That is the intention yes. Fetching the maven dependencies that can be cached should take a lot less time. Since I have never done this for a Jave project and am not familiar with maven, I can't say how effective this will be, or if there is something in this project that prevents it from helping. One way to find out though :)

mkroetzsch:
Ok, let's give it a try. Faster builds are always welcome.

mkroetzsch:
Though I find that Travis should be smart enough to transparently cache Maven dependencies locally, or at least on its own cluster, without us specifying anything about sudo, which seems quite unrelated to Maven usage. It feels like we interfere with implementation-specific mechanisms that should not be our concern; if Maven changes its download directory in the next version, we will have to reconfigure Travis?

JeroenDeDauw:
> if Maven changes its download directory in the next version, we will have to reconfigure Travis?

yeah

PR #170, opened 2015-09-04
Remove redundant modifiers

JeroenDeDauw:
Interface members can only be public, so specifying they are public is not needed.
Same for static and final.


mkroetzsch:
It is a mystery to me why this increases test coverage.

PR #172, opened 2015-09-04
Simplify boolean field assignment

JeroenDeDauw:


JeroenDeDauw:
Making builds fail when coverage % goes down is considered an anti pattern by many. % of lines covered is not what you want to focus on.

mkroetzsch:
This fail behaviour is a new feature of coveralls. It started to appear recently. Still need to find out how to turn it off. OTOH, I do spend a lot of time writing tests for other people's code. The typical pattern I see with pull requests is that they decrease coverage by 0.1%-0.5% each. You know, not testing exception handling or other corner cases. A line here, a line there. After merging a few such pulls, you are quickly down to 85% coverage, and someone (usually me) has to spend time on fixing it.

JeroenDeDauw:
I agree it's bad if code does not get tested, and that you have a problem when this starts to add up. However I really recommend to not look at the coverage %. It's very easy to get a high coverage % just by executing a lot of code. If new code that comes with tests is fully covered, that does not say anything about the quality of the tests or if they actually verify correctness of all associated production code. It's a bit like measuring the maintainability of a codebase by looking at the LOC. Both metrics are very shallow and can easily lead you to do things that are harmful, such as writing longer lines of code or tests that assert nothing but coupling that should be hidden.

PR #174, opened 2015-09-04
Use Object.equals for string comparison

JeroenDeDauw:
Before the ref equality comparator was used, presumably the code
was not working as intended.


mkroetzsch:
Thanks. Some of the code might have worked because the same string objects had been used in all cases, so that == comparison worked out. But of course this is coincidental.

PR #175, opened 2015-09-04
Remove redundant initializers

JeroenDeDauw:


wdtk-util/src/main/java/org/wikidata/wdtk/util/Timer.java
```diff
@@ -296,7 +296,7 @@ public String toString() {
 		StringBuilder labels = new StringBuilder();
 		StringBuilder values = new StringBuilder();
 
-		String separator = "";
+		String separator;
```
mkroetzsch: Why is this redundant? Java's default should be null, not "".

JeroenDeDauw: The var is always assigned to later on before it is read. Look at the if else.

PR #185, opened 2015-09-09
Use StringBuilder

JeroenDeDauw:


mkroetzsch:
I have to agree with Katie here. The code does not become more readable, and what would be the advantage? I don't think there can be any measurable performance gain here (this particular message is issued once when checking for dumps). In general, Java's compiler should be good enough to optimise short string concatenations. It's also fine to use StringBuilders, especially for longer concatenations with many runtime dependencies, but I would not replace all "+" in our code with StringBuilders now.

PR #187, opened 2015-09-11
Use false rather than always false var and camelCaps var

JeroenDeDauw:


mkroetzsch:
Note that this code is unused at the moment. It might just vanish once template constraints are moved into statements, rather than being reviewed/updated.

PR #189, opened 2015-09-11
Remove null check after method call on the object

JeroenDeDauw:
Perhaps a null check at the start of the method is needed (though not tested for)?


mkroetzsch:
Funny what kind of code a project collects over time ...

JeroenDeDauw:
CodeRot.jpg

PR #190, opened 2015-09-11
Fix presumable bug - modified collection was ignored

JeroenDeDauw:
Tests pass both before and after... guess this is not tested either


mkroetzsch:
Good catch.

PR #194, opened 2015-09-11
Fix filter test to use nonempty sitelinks (NOT DONE YET)

JeroenDeDauw:


mkroetzsch:
Well, this code was meant to be used ;-). Rather than removing it, the test should be fixed to filter only some site links and not all.

JeroenDeDauw:
As a side note: this does not happen if you test first

#RageMarkusIntoTDD

mkroetzsch:
I honestly don't know how test first should work in practice. This is not a critique -- I just cannot imagine how it would be done.

JeroenDeDauw:
If you want we can pair program a kata in Java using TDD via https://floobits.com/ (this is easy to set up)

wetneb:
Closing as stale.

PR #195, opened 2015-09-11
Remove not needed catch clause

JeroenDeDauw:


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/json/jackson/AliasesDeserializer.java
```diff
@@ -80,8 +80,6 @@
 				}
 			}
 
-		} catch (JsonProcessingException e) {
-			e.printStackTrace();
 		} catch (Exception e) {
```
mkroetzsch: I think this catch clause should be removed rather than the first. It is not good style to catch Exception, and I don't see a reason for doing this. Also, the exception handling is quite underdeveloped here. However, I believe this code is not actually used anywhere (it is meant to work around https://phabricator.wikimedia.org/T73349 but this bug affects too many places -- one would need special handwritten parsing code like this for all maps in the JSON; maybe we should just remove it)

JeroenDeDauw: Right, sure, I agree it is bad practice to catch Exception. Those changes I'm making here are done with very minimal understanding of the context, and are generally strictly refactoring (ie no behavior changes).

JeroenDeDauw: So can this be merged? It's less clutter than before and the mentioned issue was already present.

Tpt:
Let's clean it up later

PR #197, opened 2015-09-11
Remove another not needed catch clause

JeroenDeDauw:


mkroetzsch:
Actually, the better approach here would have to be not catching any exception at all. This is test helper code, and it would make sense if exceptions surface there rather than being printed. But maybe this is not crucial.

JeroenDeDauw:
Not obvious this is "test helper code"

PR #198, opened 2015-09-17
remove unused imports

mardam:
This commit removes unused imports.


mkroetzsch:
Thanks, I think these were left over during the refactoring that @JeroenDeDauw did recently.

Issue #199, opened 2015-09-20
The RDF format is incorrect


monkey2000:
Many bug trackers of RDF Databases have been report this bug.
> \"-escape sequences are not allowed in URIs, only \u

Apache Jena: https://issues.apache.org/jira/browse/JENA-806
Google Cayley: https://github.com/google/cayley/issues/256

mkroetzsch:
Thanks. We do not do any URL encoding in Wikidata Toolkit, but rely on OpenRDF for this. It seems that there are some bugs there. We ls have another issue in this space: #133

mkroetzsch:
I just checked, and eventually the URI object used to export this data is created by calling the method
```
org.openrdf.model.ValueFactory#createURI(string)
```
where the URI string is obtained as follows:
```
"http://commons.wikimedia.org/wiki/File:" + pageName.replace(' ', '_');
```
pageName here is the literal Wikipedia page name. It may contain weird characters in many cases, but no character escapes. It seems that OpenRDF accepts this input and does not throw any exception. The resulting URI object is eventually used in a triple ("statement" object in OpenRDF) and serialized ("handled") by an OenRDF Writer for RDFFormat.NTRIPLES. It really seems to be a bug in OpenRDF that this leads to malformed RDF.

monkey2000:
This bug is fixed. (As a workaround)

PR #209, opened 2015-09-26
Fix URI Format

monkey2000:
May be a solution of Issue #199


mkroetzsch:
Thanks, let's see if this helps. The branch you based this on is not master, so this is not merged into master yet. Hopefully this will happen soon, too.

monkey2000:
I wish that it will be deployed soon.

Issue #201, opened 2015-09-23
adapt label rdf representation


guenthermi:
in addition to rdfs:label export name as skos:prefLabel and schema:name.

kanikasaini:
Hi, I'm a student learning semantic web technologies and aiming to contribute to this repository. Can you please explain this issue a bit?

Thank you!

wetneb:
I am not a specialist at all, but I suspect this proposes to change the way Wikibase entities are exported to RDF: instead of just exporting labels using the `rdfs:label`, other statements using the `skos:prefLabel` and `schema:name` would be added as well.

I don't think this issue is relevant anymore since work was done to match the RDF output generated by this library to the RDF generated by Wikibase itself to populate the Query Service.

So this can probably be closed.

PR #210, opened 2015-09-28
add label representation #201

guenthermi:
This adds skos:prefLabel and schema:name properties to terms


wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/Vocabulary.java
```diff
@@ -127,6 +135,14 @@
 	}
 
 	/**
+	 * Property "name" of schema.org.
+	 */
+	public static final String SCHEMA_NAME = PREFIX_SCHEMA + "name";
+	static {
+		VOCABULARY_TYPES.put(SCHEMA_NAME, OWL_OBJECT_PROPERTY);
```
mkroetzsch: Datatype property

PR #212, opened 2015-10-08
some error handling

guenthermi:
Error handling for the case that there is no report because an action was not executed.


JeroenDeDauw:
Needs rebase

PR #215, opened 2015-10-16
Make project compatible with Java 8

julianmendez:
With these changes, the project can be compiled in Java 8. In addition, the Travis CI configuration file is updated to ensure compatibility.


mkroetzsch:
Very good, thanks!

Issue #217, opened 2015-10-29
NulPointeException API Connection


maatinito:
Following the report on stack overflow
Using the sample line
```java 
Map<String, EntityDocument> results = wbdf.getEntityDocuments("Q80", "P31");
```
Triggers and exception
```java
    Exception in thread "main" java.lang.NullPointerException
        at org.wikidata.wdtk.wikibaseapi.ApiConnection.fillCookies(ApiConnection.java:544)
        at org.wikidata.wdtk.wikibaseapi.ApiConnection.sendRequest(ApiConnection.java:346)
        at org.wikidata.wdtk.wikibaseapi.WbGetEntitiesAction.wbGetEntities(WbGetEntitiesAction.java:187)
```
Code:
```java 
		List<String> headerCookies = headerFields
				.get(ApiConnection.HEADER_FIELD_SET_COOKIE);
		for (String cookie : headerCookies) {
```

headerCookies is null :-(

See also the question on Stack Overflow : http://stackoverflow.com/questions/33378950/wikidata-api-always-returns-a-null-pointer-exception

guenthermi:
Thank you for the report.

PR #218, opened 2015-10-29
add null check for headerCookies in fillCookies

guenthermi:
This should fix the bug according to issue #217. If the header field Set-Cookie was not used headerCookies got null which caused a NullPointerException.


JeroenDeDauw:
+1. @mkroetzsch not sure if I'm allowed to merge this or not :)

guenthermi:
I think I just do it myself

Issue #151, opened 2015-06-28
Follow HTTP redirects


mkroetzsch:
The current code in WebResourceFetcher does not follow HTTP redirects. This causes problems, for example, if somebody uses an http-based URL http://www.wikidata.org/w/api.php to access the Wikidata API. This URL redirects to https://www.wikidata.org/w/api.php, but WDTK fails to follow this and encounters an error.

This is probably a minor issue since one can usually use the proper URL directly on Web requests, but it makes all online functions of WDTK somewhat less robust. The WebResourceFetcher should better check for redirects (and maybe for other types of non-200 responses) and act appropriately.

wetneb:
This has probably been solved by #507 but it would be worth double-checking.

PR #221, opened 2015-11-15
Follow HTTP Redirects

monkey2000:
Fix for Issue #151 
Add a test case for redirecting.

---

BTW: I don't think that `getInputStreamForUrl` is a instance method.


JeroenDeDauw:
Doing this cast changes the behaviour of the code?

monkey2000:
Yes. Using `setFollowRedirects` and I'm sure that it will handle redirects.
The test case passed.

JeroenDeDauw:
Oh, I had not seen the `urlConnection.setFollowRedirects(true);` line. Silly me :)

> The test case passed.

Because you wrote it like that, I started wondering if you also verified it failed before and decided to find out. I can undo your changed to the production code, and the added test will still pass.

monkey2000:
The test case is copyed from issue #151.
Seems like it can't handle redirects before.

PR #229, opened 2016-03-04
Wbsearchentities

sbrunk:
I did an implementation of the [wbsearchentities](https://www.wikidata.org/w/api.php?action=help&modules=wbsearchentities) API call as discussed #228.
I wasn't sure about the best place to put the search result structure (WbSearchEntitiesResult) and its implementation for deserialization (JacksonWbSearchEntitiesResult) since it is not really part of the coredatamodel. So for now it's in the wikibaseapi package.
No integration in WikibaseDataFetcher yet.


Tpt:
Thank you very much for the pull request. I have merged #257 that is based on yours

Issue #234, opened 2016-06-17
JSON deserialization exception on daily dumps


Tpt:
On the last daily dump (wikidatawiki-20160616-pages-meta-hist-incr.xml.bz2) I have the following Jackson error:

```
2016-06-17 19:54:13 ERROR - Failed to map JSON for item Q41240: Can not deserialize instance of java.util.LinkedHashMap out of START_ARRAY token
[...]

PR #235, opened 2016-06-17
Do not fail when descriptions/sitelinks are serialized as "[]" (empty array)

Tpt:
Based on the DeserializationFeature.ACCEPT_EMPTY_ARRAY_AS_NULL_OBJECT option of Jackson

Tries to work around #234


JeroenDeDauw:
This bug is still there / there again in wd? >_>

Tpt:
Yes, sadly :-(

Bug on Phabricator: https://phabricator.wikimedia.org/T138104

mkroetzsch:
> Based on the DeserializationFeature.ACCEPT_EMPTY_ARRAY_AS_NULL_OBJECT option of Jackson

We looked before but had not found this option! Thanks a lot, if this works it is the best solution.

mkroetzsch:
P.S. this adds to my determination to create a new release soon ...

PR #241, opened 2016-09-08
Do not fail when reading redirects

### LINKED ISSUE #240: Deserialization failure on dumps with redirects
The Wikidata daily XML dumps contains serialization for redirects.

Wikidata Toolkit 0.7 fails when deserializing them:

``` java
GRAVE: Failed to map JSON for item Q26790101: Unexpected token (END_OBJECT), expected FIELD_NAME: missing property 'type' that is to contain type id  (for class org.wikidata.wdtk.datamodel.json.jackson.JacksonItemDocument)
 at [Source: {"entity":"Q26790101","redirect":"Q21863925"}; line: 1, column: 45]
com.fasterxml.jackson.databind.JsonMappingException: Unexpected token (END_OBJECT), expected FIELD_NAME: missing property 'type' that is to contain type id  (for class org.wikidata.wdtk.datamodel.json.jackson.JacksonItemDocument)
 at [Source: {"entity":"Q26790101","redirect":"Q21863925"}; line: 1, column: 45]
    at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:148)
    at com.fasterxml.jackson.databind.DeserializationContext.wrongTokenException(DeserializationContext.java:927)
    at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer._deserializeTypedUsingDefaultImpl(AsPropertyTypeDeserializer.java:151)
    at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer.deserializeTypedFromObject(AsPropertyTypeDeserializer.java:103)
    at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeWithType(BeanDeserializerBase.java:966)
    at com.fasterxml.jackson.databind.deser.impl.TypeWrappedDeserializer.deserialize(TypeWrappedDeserializer.java:42)
    at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:1443)
    at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1081)
    at org.wikidata.wdtk.dumpfiles.WikibaseRevisionProcessor.readValue(WikibaseRevisionProcessor.java:160)
    at org.wikidata.wdtk.dumpfiles.WikibaseRevisionProcessor.processItemRevision(WikibaseRevisionProcessor.java:97)
    at org.wikidata.wdtk.dumpfiles.WikibaseRevisionProcessor.processRevision(WikibaseRevisionProcessor.java:88)
    at org.wikidata.wdtk.dumpfiles.MwRevisionProcessorBroker.notifyMwRevisionProcessors(MwRevisionProcessorBroker.java:183)
    at org.wikidata.wdtk.dumpfiles.MwRevisionProcessorBroker.processRevision(MwRevisionProcessorBroker.java:147)
    at org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor.processXmlRevision(MwRevisionDumpFileProcessor.java:427)
    at org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor.processXmlPage(MwRevisionDumpFileProcessor.java:345)
    at org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor.tryProcessXmlPage(MwRevisionDumpFileProcessor.java:282)
    at org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor.processXmlMediawiki(MwRevisionDumpFileProcessor.java:202)
    at org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor.processDumpFileContents(MwRevisionDumpFileProcessor.java:155)
    at org.wikidata.wdtk.dumpfiles.DumpProcessingController.processDumpFile(DumpProcessingController.java:546)
    at org.wikidata.wdtk.dumpfiles.DumpProcessingController.processAllRecentRevisionDumps(DumpProcessingController.java:389)
```

Tpt:
Closes #240

### DIFF
--- a/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/WikibaseRevisionProcessor.java
+++ b/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/WikibaseRevisionProcessor.java
@@ -47,7 +47,6 @@ public class WikibaseRevisionProcessor implements MwRevisionProcessor {
 
 	static final Logger logger = LoggerFactory
 			.getLogger(WikibaseRevisionProcessor.class);
-	static final Pattern redirectionsPattern = Pattern.compile("\\{\"entity\":\"[^\"]*\",\"redirect\":\"[^\"]*\"\\}");
 
 	/**
 	 * The IRI of the site that this data comes from. This cannot be extracted
@@ -165,7 +164,7 @@ public class WikibaseRevisionProcessor implements MwRevisionProcessor {
 	}
 
 	private boolean isWikibaseRedirection(MwRevision mwRevision) {
-		return redirectionsPattern.matcher(mwRevision.getText()).matches();
+		return mwRevision.getText().contains("\"redirect\":"); //Hacky but fast
 	}
 
 	public <T> T readValue(String content, Class<T> valueType) throws IOException {
--- a/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/WikibaseRevisionProcessor.java
+++ b/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/WikibaseRevisionProcessor.java
@@ -22,6 +22,7 @@ package org.wikidata.wdtk.dumpfiles;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import org.slf4j.Logger;
@@ -46,6 +47,7 @@ public class WikibaseRevisionProcessor implements MwRevisionProcessor {
 
 	static final Logger logger = LoggerFactory
 			.getLogger(WikibaseRevisionProcessor.class);
+	static final Pattern redirectionsPattern = Pattern.compile("\\{\"entity\":\"[^\"]*\",\"redirect\":\"[^\"]*\"\\}");
 
 	/**
 	 * The IRI of the site that this data comes from. This cannot be extracted
@@ -93,6 +95,10 @@ public class WikibaseRevisionProcessor implements MwRevisionProcessor {
 	}
 
 	public void processItemRevision(MwRevision mwRevision) {
+		if(isWikibaseRedirection(mwRevision)) {
+			return;
+		}
+
 		try {
 			JacksonItemDocument document = readValue(mwRevision.getText(), JacksonItemDocument.class);
 			document.setSiteIri(this.siteIri);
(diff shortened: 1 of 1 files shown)


mkroetzsch:
Thanks. Which dumps may actually contain redirects now? JSON or XML or both?

Tpt:
Currently only JSON

wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/WikibaseRevisionProcessor.java
```diff
@@ -154,6 +164,10 @@ public void processPropertyRevision(MwRevision mwRevision) {
 
 	}
 
+	private boolean isWikibaseRedirection(MwRevision mwRevision) {
+		return redirectionsPattern.matcher(mwRevision.getText()).matches();
```
thiemowmde: Given that one of the main goals of this code is to be as fast as possible, I believe this should not use a regular expression but a `startsWith` or something similar.

Tpt: Done with `.contains` and string `"redirect":` in order to don't detect other structures starting with `{"entity":` wrongly as redirects if they are introduced in the future.

Tpt:
@mkroetzsch would be great to have this merged before the next release

thiemowmde:
I really don't know if the substring matching approach currently in this patch is the best thing to do. Note that this will fail whenever an element *anywhere* in an entity contains a "redirect" array key. Currently this can never happen, but may happen any time when new entity types or data value types are introduced.

Isn't it possible to catch a specific exception when one of the Jackson…Documents is created? This should fail when fed with the revision text of a redirect, I believe.

Tpt:
> Currently this can never happen, but may happen any time when new entity types or data value types are introduced.

Yes, definitely. But the introduction of new entity types or data value types will break the library anyway (Wikidata dump reading is in fact currently broken on all entities with QuantityValues with null upper/lower values).

> Isn't it possible to catch a specific exception when one of the Jackson…Documents is created?

Yes, we could catch the Jackson deserialization exception but we would have to filter if the exception is raised because of a redirect or because of an other unsupported Wikibase feature (new value types...).

A clean implementation would probably be to add support of redirects to the deserializer itself (with an EntityRedirect type...). I mostly see this PR as a quick fix.

Tpt:
Merging this change to have a working implementation with daily dumps. We should definitely have an interface for redirects in the future.

PR #244, opened 2016-09-24
Disables XML secure processing for MW dump files

### LINKED ISSUE #243: XMLStreamException exception when parsing revision dumps
When parsing some big revision dumps this error is raised:

```
sept. 23, 2016 7:05:46 PM org.wikidata.wdtk.dumpfiles.MwRevisionDumpFileProcessor processDumpFileContents
SEVERE: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[291117,14066]
Message: JAXP00010004: The accumulated size of entities is "50 000 001" that exceeded the "50 000 000" limit set by "FEATURE_SECURE_PROCESSING".
```

It is probably because the XML files contains to many nodes with respect to the limit set by XML secure processing. We should probably set XMLConstants.FEATURE_SECURE_PROCESSING option to false as Wikimedia dump should be trustable.

Tpt:
These file are probably trustable and it allows to parse all files, even huge ones.

Closes #243

### DIFF
--- a/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/MwRevisionDumpFileProcessor.java
+++ b/wdtk-dumpfiles/src/main/java/org/wikidata/wdtk/dumpfiles/MwRevisionDumpFileProcessor.java
@@ -24,6 +24,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import javax.xml.XMLConstants;
 import javax.xml.stream.XMLInputFactory;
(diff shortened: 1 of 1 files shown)


mkroetzsch:
There seems to be a problem with this. The test fail due to 

```java.lang.IllegalArgumentException: Property http://javax.xml.XMLConstants/feature/secure-processing is not supported```

Would we need to use another version of Java or some library for this to work?

Tpt:
It seems that XMLInputFactory does not supports this property even if the error reported in #243 happens with Oracle JDK 8 on Mac/Linux

PR #247, opened 2016-11-14
Makes QuantityValue bounds optional

Tpt:
Closes #246

PR #248, opened 2016-12-15
Ignore null property types when generating RDF.

elshimone:
This fixes an issue I found after applying the changes to make QuantityValue bounds optional (https://github.com/Wikidata/Wikidata-Toolkit/pull/247).

Exporting a wikidata dump as rdf triggered a null pointer exception:

Exception in thread "main" java.lang.NullPointerException
        at org.wikidata.wdtk.rdf.SnakRdfConverter.getRangeUri(SnakRdfConverter.java:307)

It looks like the intention of the code was to return null in the case of an unknown property type, this informed my fix.

### DIFF
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/SnakRdfConverter.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/SnakRdfConverter.java
@@ -304,8 +304,9 @@ public class SnakRdfConverter implements SnakVisitor<Void> {
 		String datatype = this.propertyRegister
 				.getPropertyType(propertyIdValue);
 
-        if (datatype == null)
-            return null;
+		if (datatype == null)
+			return null;
+
 		switch (datatype) {
 		case DatatypeIdValue.DT_STRING:
 		case DatatypeIdValue.DT_EXTERNAL_ID:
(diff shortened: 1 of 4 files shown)


wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/SnakRdfConverter.java
```diff
@@ -304,6 +304,9 @@ String getRangeUri(PropertyIdValue propertyIdValue) {
 		String datatype = this.propertyRegister
 				.getPropertyType(propertyIdValue);
 
+		if (datatype == null)
+			return null;
```
thiemowmde: I believe the `@return` above should be adjusted.

wdtk-rdf/src/test/java/org/wikidata/wdtk/rdf/MockPropertyRegister.java
```diff
@@ -47,6 +47,12 @@ public MockPropertyRegister() {
 
 	}
 
+	public MockPropertyRegister (String uriPatternPropertyId,
+								 ApiConnection apiConnection, String siteUri) {
+		super(uriPatternPropertyId, apiConnection, siteUri);
```
thiemowmde: Why is this necessary if it doesn't change the signature anyway?

guenthermi:
Thank you! I just removed the second constructor and merged the Code via #249.

PR #259, opened 2017-07-04
Adds OpenJDK 8 to Travis configuration

Tpt:


Tpt:
Travis version of Trusty is not able to run OracleJDK7 with sudo: false. I abandon this change for now.

Issue #271, opened 2017-09-13
Unable to login: NullPointerException


wetneb:
Logging in with this library can fail with a NullPointerException.

Here is my code:
<pre>
import org.wikidata.wdtk.wikibaseapi.ApiConnection;
import org.wikidata.wdtk.wikibaseapi.LoginFailedException;

public class WikidataTest {
	public static void main(String [] args) {
		ApiConnection conn = ApiConnection.getWikidataApiConnection();
		try {
			conn.login("this is a valid username", "corresponding password");
		} catch (LoginFailedException e) {
			e.printStackTrace();
		}
	}
}
</pre>

I obtain the following trace:
<pre>
Exception in thread "main" java.lang.NullPointerException
	at java.net.URLEncoder.encode(URLEncoder.java:205)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.getQueryString(ApiConnection.java:602)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.sendRequest(ApiConnection.java:330)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.confirmLogin(ApiConnection.java:478)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.login(ApiConnection.java:258)
	at WikidataTest.main(WikidataTest.java:9)
</pre>

wetneb:
So, after hours (sigh) of investigation, it looks like the POST requests made by the toolkit are not well-formed enough. Pointing the toolkit to a local server that replays the requests to the Wikidata API fixes the bug. It looks like this problem could be solved by using a higher-level HTTP Java library which would take care of the heavy-lifting in a clean way. That would add a dependency to the toolkit, but hey… it's currently unusable to perform any edits!

PolyglotOpenstreetmap:
Could it be that this fix is not in wikidata-toolkit-with-dependencies-0.7.0.jar, available for download? I'm using the jar file from Jython, but I can't log in.

wetneb:
@PolyglotOpenstreetmap yes this fix has not been released yet… I guess it is quite urgent to make a release. @Tpt let me know if I can help with that!

PolyglotOpenstreetmap:
I don't understand why this issue was closed. I downloaded the jar again and it's still giving the same error message.

wetneb:
@PolyglotOpenstreetmap have you tried building the toolkit from source? downloading released jars again will not solve your problem.

[...]

PR #272, opened 2017-09-13
Debugs the API calls library

Tpt:


wetneb:
Thanks a lot, it does solve #271 indeed!

PR #275, opened 2017-10-06
Serialize StatementUpdate with Jackson.

### LINKED ISSUE #274: Extending StatementUpdate
I am trying to extend [StatementUpdate](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java#L47) to support updating terms too (labels, descriptions and aliases). The simplest way to do that as a user seems to be to inherit from that class and add the required functionality on top of that.

However, I run into an issue when generating the JSON of the change. Naturally, I would expect to just call [`StatementUpdate.getJsonUpdateString`](https://github.com/Wikidata/Wikidata-Toolkit/blob/3541b49e0961af0d4e754e64904d06e29766f684/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java#L100-L130) and extend the JSON payload it returns. I was horrified to discover that its implementation generates JSON manually, with a StringBuilder… I am not a Java wizard, but surely there must be better ways to generate JSON than that? I would expect the conversion to String to happen much later downstream (like, just before the HTTP request is performed, ideally), which would make it possible to extend the JSON object easily.

So, what should I do?
- Keep my current approach, just call `StatementUpdate.getJsonUpdateString`, remove the last character (`}`) of the JSON payload and append the rest of the content of the JSON object with a StringBuilder. As far as I can tell this is a nasty hack.
- Parse the String returned by `StatementUpdate.getJsonUpdateString` with some library, add my labels, descriptions and aliases cleanly via this library, and convert the result back to a String (to preserve consistency of the interface with `StatementUpdate`). That's a bit cleaner but it feels really wrong.
- Rewrite `StatementUpdate` (and its users) so that it exposes a cleaner interface (so, make a PR). That's a lot more work. What library should I use? Jackson? Org.json?

wetneb:
This makes the class more reusable. Closes #274.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
@@ -28,6 +28,7 @@ import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
+import org.wikidata.wdtk.datamodel.helpers.DatamodelConverter;
 import org.wikidata.wdtk.datamodel.interfaces.Claim;
 import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
 import org.wikidata.wdtk.datamodel.interfaces.Reference;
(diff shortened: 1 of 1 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
```diff
@@ -65,8 +72,61 @@ public StatementWithUpdate(Statement statement, boolean write) {
...
 		}
 	}
+	
+	/**
+	 * Helper class to ease serialization of deleted statements. Jackson will
+	 * serialize this class according to the format required by the API.
+	 * 
+	 * @author antonin
+	 */
+	class DeletedStatement implements Statement {
```
Tpt: Not sure why it is relevant to implement Statement here (but maybe it is)

wetneb: @Tpt it is currently needed, because I push instances of that to a `List<Statement>` later (see https://github.com/Wikidata/Wikidata-Toolkit/pull/275/files#diff-32894a9bb9d9ca9fd81765c38117eb36R185). I found that quite neat but I'm happy to consider other solutions!

Tpt: Ok! Thanks!

PR #276, opened 2017-10-06
Add support for updating terms on items.

wetneb:
Technically not all the operations exposed by the API are supported yet: it is not possible to delete labels and descriptions yet. I am not sure I want to add support for that as I don't think it is critical.

This PR is not quite finished yet as it would be good to expose this functionality in WikibaseDataEditor. But if we want to move to a Builder-based construction, we should adapt WikibaseDataEditor accordingly - I am not sure what it would look like.

I am also not very happy with the fact that it currently only accepts `ItemDocument`s as initial entity - it should accept `PropertyDocument`s too, but it looks like curiously there is no common superclass of these two, so generics seem to be required here?

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
@@ -42,6 +42,8 @@ import org.wikidata.wdtk.datamodel.json.jackson.JacksonObjectFactory;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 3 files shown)


Tpt:
Thank you!

> I am also not very happy with the fact that it currently only accepts ItemDocuments as initial entity - it should accept PropertyDocuments too, but it looks like curiously there is no common superclass of these two, so generics seem to be required here?

The datamodel is done by compositing some interfaces (EntityDocument, TermedDocument, StatementDocument...) and implementations. It's going to be very useful when we will add support of new kinds of entities like lexemes (that are going to have Statements too but not the same kind of terms (e.g. it does not makes sense to translate a lexeme)).
For an example see https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/ItemDocumentImpl.java

A way to deal with that would be to have StatementUpdate, TermUpdate, XYZUpdate as separated classes that contains most of the logics and build ItemUpdate, PropertyUpdate, LexemeUpdate... that are composed using these classes and provides the user-facing API (we would have a lot of 1 line methods that just call the relevant method of a composed class...).

wetneb:
Okay! I don't know how to do this mixing in Java (it's essentially an imitation of multiple inheritance, right?). Feel free to migrate TermStatementUpdate to TermUpdate accordingly!

PR #308, opened 2018-01-13
Skip unintentional null edits

wetneb:
If the changes made on an item end up being spurious (adding statements that already exist, deleting statements that do not exist, setting labels to their current value, and so on), then by default we do not want to perform any edit (because it costs time, bandwidth, server resources for the Wikibase instance).

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
@@ -189,6 +189,16 @@ public class StatementUpdate {
 		}
 		return updatedStatements;
 	}
+	
+	/**
+	 * Returns true when the edit is not going to change anything on the item.
+	 * In this case, the change can be safely skipped, except if the side effects
+	 * of a null edit are desired.
+	 */
+	@JsonIgnore
+	public boolean isNull() {
+		return getUpdatedStatements().isEmpty();
+	}
 
 	/**
 	 * Marks the given lists of statements for being added to or deleted from
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
@@ -196,7 +196,7 @@ public class StatementUpdate {
 	 * of a null edit are desired.
 	 */
 	@JsonIgnore
-	public boolean isNull() {
+	public boolean isEmptyEdit() {
 		return getUpdatedStatements().isEmpty();
 	}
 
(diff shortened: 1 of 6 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
```diff
@@ -189,6 +189,16 @@ public String getJsonUpdateString() {
...
 		return updatedStatements;
 	}
+	
+	/**
+	 * Returns true when the edit is not going to change anything on the item.
+	 * In this case, the change can be safely skipped, except if the side effects
+	 * of a null edit are desired.
+	 */
+	@JsonIgnore
+	public boolean isNull() {
```
Tpt: A method name like "isNullEdit" or "isEmptyEdit" is maybe more meaningful. Here it looks a lot like a method to check if something == null.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java
```diff
@@ -551,10 +556,84 @@ public ItemDocument updateTermsStatements(ItemDocument currentDocument,
...
+	}
+	
+	/**
+	 * Performs a null edit on an item. This has some effects on Wikibase,
+	 * such as refreshing the labels of the referred items in the UI.
+	 * 
+	 * @param currentDocument
+	 * 			the document to perform a null edit on
+	 * @param summary
+	 * 			a summary that can be passed to Wikibase (although it is not
```
Tpt: If it is a real null edit the summary is ignored because no new revision is created in the database. I would be in favor of removing this argument.

wdtk-wikibaseapi/src/test/java/org/wikidata/wdtk/wikibaseapi/TermStatementUpdateTest.java
```diff
@@ -119,6 +121,7 @@ public void testAddLabelAndAlias() throws JsonProcessingException {
 		assertEquals(Collections.singleton("fr"), su.getAliasUpdates().keySet());
 		assertEquals(alias.getText(), su.getAliasUpdates().get("fr").get(0).getText());
 		assertTrue(su.getDescriptionUpdates().isEmpty());
+		assertTrue(!su.isNull());
```
Tpt: assertFalse would be more readable.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/StatementUpdate.java
wetneb: makes sense, I will change that.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java
wetneb: great, I wasn't sure about this, so let's do that.

wdtk-wikibaseapi/src/test/java/org/wikidata/wdtk/wikibaseapi/TermStatementUpdateTest.java
wetneb: for some reason Eclipse was telling me it didn't exist (not sure why though)

Tpt:
Let's merge this! I would slightly prefer "isNullEdit" instead of "isEmptyEdit" (because it's how it is called usually) but it could be done later.

PR #312, opened 2018-01-17
Datamodel code cleanup

Tpt:


Tpt:
It's mostly cosmetic changed done using IDE tools so let's close it and do it again on top of #312

PR #316, opened 2018-01-23
[DO NOT MERGE] Granular editing

wetneb:
This is a work in progress and should not be merged yet. I'm creating this PR to discuss how to modify existing instances of EntityDocuments. The solution in #315 is not enough, because it only lets us modify existing PropertyDocument or ItemDocument. The issue is that most editing actions will be performed on a StatementDocument or TermedDocument depending on the change, so we need a way to modify these too (without casting them to their subclasses). Commit https://github.com/Wikidata/Wikidata-Toolkit/commit/a61c6f8b81c016c762f707fa3b095c386e68270d introduces a (very ugly) EntityDocumentBuilderFactory that helps with that. Is this the right solution? If not, how to modify instances of TermedDocument or StatementDocument (as demonstrated in the tests)?

I am starting to think that these EntityDocument interfaces should just have public setters. Having completely unmodifiable entity documents made sense as long as only wbeditentity was used, but the bureaucracy it induces for other operations looks quite daunting. The risk with that is that users use these setters and expect edits to be performed. But on the other hand, once an edit is performed, we probably don't want the old EntityDocument to persist in memory: we could want this instance to be updated by the new contents of the entity.


Tpt:
> Is this the right solution?
I am not sure. It is quite hugly.

> I  am starting to think that these EntityDocument interfaces should just  have public setters.

I still believe that having immutable EntityDocument would be better to keep things safe.

But we could add methods like:

```
interface TermedDocument {
    TermedDocument withLabel(MonolingualTextValue v); //Returns a new instance of the TermedDocument with the label
}
```

The problem with this solution is that we loose the actual concrete type (Item or Property) of return.

We could instead do:
```
interface TermedDocument<T extend TermedDocument> {
    T withLabel(MonolingualTextValue v); //Returns a new instance of the TermedDocument with the label
}

interface Item extends TermedDocument<Item> {
}
```
but it's a bad breaking change. I am not sure which solution is best.
We could maybe have EditableX interfaces that each extend X with X EntityDocument, Item, TermedDocument... It will allow to not break the editing interface and not have to implement these methods in the JSON implementation. We would then change DatamodelConverter and Factory to allow to build instances of these interfaces. But it's quite heavy.

wetneb:
@Tpt I like the idea of your last solution, but note that DatamodelConverter does not have any methods to convert TermedDocuments or StatementDocuments yet - so that is also something we would need to introduce (and it would also introduce casts like the ones in EntityDocumentBuilderFactory in the existing converters). When we will introduce LexemeDocument, all these casts will need to be updated…

Other solutions:
* give up the requirement to return the new value of the entity after a change
* use the Jackson models in the API wrapper. Arguably it's not that of a breach if it happens in a context where we are dealing with JSON anyway (we are parsing the server response).
* tweak the JSON response from the server to transform it into the desired document before unmarshalling it into an instance

But I think that the need to modify instances will probably bite us back later on, so these are not long term solutions…

mkroetzsch:
Data objects should be immutable. If they were modifiable, then you could not store them in hashmaps, since their physical location in the data structure depends on their hash (based on their content). If you modify an object that is also stored in a hashmap anywhere in your code, then you will break the contract of that data structure (leading to unpredictable behaviour of the hashmap, such as holding two distinct copies of the same data or not being able to retrieve objects properly).

More generally, immutable objects can safely be shared across your code. You can copy the reference instead of creating a new instance each time you pass an object as a parameter. If objects are modified by some parts of the code, it is very likely that you will find other parts of the code that have used the same reference to be very surprised by this. The hashmap case is a special example of this. I know of large (commercial) code bases have been struggling with almost untraceable errors after giving in to the idea of making objects "conveniently" mutable. It is not convenient. It is saving some keystrokes now at the price of hours of frustrating debugging time later.

The only way around these problems would be to have a content-independent identity for all mutable objects, so their hash and equals methods are unrelated to their actual content. This is, in my view, very unnatural and undesirable, since two documents with exactly the same contents would then not be equal any more. Immutable objects are the right software engineering solution here.

If existing factories are not convenient enough, then suitable builder patterns can make object creation more pleasant. In particular, a builder can be mutable as long as the object was not really created yet, so this would be easy to do.

wetneb:
@mkroetzsch okay, it all makes sense. So, how would you go around this problem of creating modified copies of TermedDocument or StatementDocument instances? @Tpt and I have proposed various solutions but it's not clear to me which one is best.

wetneb:
More specifically, I'd be very happy with a suitable builder for that - the issue is that I don't see how we could make a TermedDocumentBuilder as TermedDocument is an interface that is not meant to be implemented directly (but only through specializations such as ItemDocument or PropertyDocument).

Maybe the requirement to create modified copies of instances of TermedDocument is just wrong in the first place for reasons I am not aware of. In this case, let me explain again the situation where I need this. I want the editing interface to select the most specific Wikibase API action that can be used to perform a StatementUpdate (mainly so that the edit summaries are more meaningful, and localized server-side). Currently, when performing a StatementUpdate, the new version of the document is returned: this is simply done by deserializing the server's response which contains the full updated document. The more specific actions (such as wbsetclaim) do not return the whole document (but for instance only the statement that was modified), so if we want to preserve this behavior I need a way to create a copy of the original document updated with the appropriate changes (revision id and new statement for instance). As StatementUpdate works on StatementDocument instances (which looks sensible to me), I need a way to create a modified copy of these, ideally without casting them to their more specific interfaces (ItemDocument and PropertyDocument).

So I'm all in for established OOP guidelines - I just need a concrete solution for this problem.

mkroetzsch:
Yes, I understand your problem. A builder could work as follows. It would store an internal, mutable copy of all data needed to create a fresh instance of such a document, and it would have methods to modify any of this data as desired. You can create an actual instance of this at any time, but when doing so you have to provide a factory to the builder. This will determine which implementation the builder will use. Alternatively, the builder could receive such a factory on create, or make its own factory internally as a default.

I think this scheme should be reasonably convenient when you only want to modify "top-level" aspects of the document object. If you want to make changes to its sub-objects (e.g., modify a statement rather than just add or remove statements), then you still need to create new objects for those immutable subobjects (or have similar builder patterns there, but this only pays of for complex objects where the direct call to the factory method would be very lengthy).

wetneb:
I might be wrong but I think we are just over-thinking this. As @Tpt suggests, we can add methods like these to `TermedDocument`:

    public TermedDocument withLabel(MonolingualTextValue label);

and in `ItemDocument`:

    @Override
    public ItemDocument withLabel(MonolingualTextValue label);

we don't need any generics for that because `ItemDocument` is a subclass of `TermedDocument`.

This seems completely non-intrusive, sane and simple to implement.

wetneb:
Unless you have any concern about this solution, I will follow that route in a new PR.

PR #320, opened 2018-02-08
Revamp the datamodel

wetneb:
<s>This is a work in progress, I am aware of a few things that are still missing:</s>
* [x] wrapping fields in unmodifiable containers where appropriate
* [x] splashing my name all over the place with `@author`
* [x] maybe reusing the names of the deleted package for the Jackson package? So that the classes are clearly identifiable as the default implementations provided by the library and to spare users the need to replace explicit mentions of these in their code (but in theory they should not have any - they should just use the `Datamodel` helper, which has not changed).
* [x] migrate the deleted tests to the new implementation (I deleted them without realizing that they were not replicated in the jackson package)

I've left a few comments in the commit messages about some important changes.

Happy reviewing. 😛


Tpt:
Code changes were looking good to me. Let's improve the codebase in other pull requests

PR #322, opened 2018-02-20
Make ApiConnection serializable with Jackson

### LINKED ISSUE #321: Serializing ApiConnection
I found myself needing to serialize `ApiConnection`, so that I can keep a user logged in across multiple runs without storing their password in clear text. The password field should not be serialized, only the cookies should be remembered.

wetneb:
Closes  #321.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -208,8 +208,11 @@ public class ApiConnection {
 	 * Deserializes an existing ApiConnection from JSON.
 	 * 
 	 * @param apiBaseUrl
+	 * 		base URL of the API to use, e.g. "https://www.wikidata.org/w/api.php/"
 	 * @param cookies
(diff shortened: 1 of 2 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -194,6 +201,29 @@
...
+	 * 
+	 * @param apiBaseUrl
+	 * 		base URL of the API to use, e.g. "https://www.wikidata.org/w/api.php/"
+	 * @param cookies
+	 * 		map of cookies used for this session
+	 * @param loggedIn
+	 * 		true if login succeeded.
+	 */
+	@JsonCreator
+	protected ApiConnection(
```
Tpt: What about making it package-private? It seems cleaner than protected.

wetneb: Why not! When refactoring the datamodel I have often found myself reusing these constructors at other places in the package so I generally mark these as protected, but I don't have this need in this particular case.

PR #323, opened 2018-02-23
Custom editing actions

wetneb:
This adds a few "with" methods to various interfaces of the datamodel, implements and tests them in the default implementation, and uses them to change the editing interface to choose more granular API actions depending on the nature of the change being pushed to Wikidata.


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/EntityDocument.java
```diff
@@ -46,5 +46,10 @@
 	 * @return revision id
 	 */
 	long getRevisionId();
+	
+	/**
+	 * Returns a copy of this document with an updated revision id.
+	 */
+	EntityDocument withRevisionId(long newRevisionId);
```
Tpt: What is the use case of this setter? It seems quite dangerous to allow setting the revisionId outside of the deserialization of content retrieved from Wikibase. It is maybe useful when building the result of an edit action but we should at least add a strong warning to this method.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Statement.java
```diff
@@ -89,4 +89,11 @@
 	 * @return main value of the statement, or null
 	 */
 	Value getValue();
+	
+	/**
+	 * Returns the same statement, but with a different identifier.
+	 * This is useful when the existing identifier was empty and we need to 
+	 * attribute one before creating the statement in a remote Wikibase instance.
+	 */
+	Statement withStatementId(String id);
```
Tpt: It would be great to avoid this method and assign a statement id as soon a statement object is created. But I believe that in this case we would have troubles using the wbcreateclaim method that returns a new statementId for the newly created claim. There is maybe a nicer middle ground (but I don't see one yet).

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * This interface just joins {@link TermedDocument} and {@link StatementDocument}.
+ * 
+ * It is necessary to introduce this interface because the conflict between 
+ * the return types of the withRevisionId method in both interfaces needs to be resolved.
+ * @author antonin
+ *
+ */
+public interface TermedStatementDocument extends TermedDocument, StatementDocument {
```
Tpt: I would avoid to introduce this interface because it will introduce problems when we will want to implement support of lexemes that are going to be StatementDocument but not TermedDocument. A possible way, if we do not manage to get free of the withRevisionId method, would be to avoid overriding it in the TermedDocument and StatementDocument interfaces.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
```diff
@@ -123,9 +125,10 @@
 	 *            prefix of entity URIs), e.g.,
 	 *            "http://www.wikidata.org/entity/"
 	 */
-	public WbEditEntityAction(ApiConnection connection, String siteIri) {
+
+	public WbEditingAction(ApiConnection connection, String siteUri) {
 		this.connection = connection;
-		this.siteIri = siteIri;
+		this.siteIri = siteUri;
```
Tpt: Why this renaming?

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
```diff
@@ -287,8 +290,379 @@ public EntityDocument wbEditEntity(String id, String site, String title,
...
+	 * @param newEntity
+	 *            used for creating a new entity of a given type; the value
+	 *            indicates the intended entity type; possible values include
+	 *            "item" and "property"; if used, the parameters id, site, and
+	 *            title must be null
+	 * @param language
+	 *            the language code for the label
+	 * @param value
+	 *            the value of the label to set. Set it to null to remove the label.
+     * @param bot
```
Tpt: Typo

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
```diff
@@ -287,8 +290,379 @@ public EntityDocument wbEditEntity(String id, String site, String title,
...
```
Tpt: Typo

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/EntityDocument.java
wetneb: Yes - it is not meant to be used by library users, but I haven't found a better solution. Here is the use case:
* you have an EntityDocument
* you edit some part of it (add an alias, say) with the API
* the server returns the new revid and the new list of aliases for that language
* we need to construct the new version of the document, which is: the initial state of the document, with the updated aliases and the updated revision id. Hence `withRevisionId` and `withAliases`.

One thing we could do is merge the two methods: each `with` method of `EntityDocument`, `StatementDocument`, etc would also take the new revision id as a second parameter. This way, there would be no natural way to change just the revision id, or just the content.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Statement.java
wetneb: The problem with that is that you still need to be able to represent "statements that do not exist on the wikibase server yet"… If you give them an ID you will not be able to distinguish them from the existing ones. I think we rely on this distinction in StatementUpdate (but I am not entirely sure).

One other issue with assigning ids at creation is that it relies on the random number generator, so it might make things a bit more complicated to test (or you need to pass on a GuidGenerator in many classes…) but I guess that can be overcome.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
wetneb: In the current code we do need to override it in `TermedDocument` and `StatementDocument` (if we want `StatementUpdate` to return modified `StatementDocument` instances).

It's not clear to me why this interface is an issue - the lexemes are just not going to implement it, and that's fine (they will not be usable in the contexts where this interface is required). Classes like `StatementUpdate` should still be able to work on them.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
wetneb: That's inherited from the Jackson implementation - I will fix this.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
Tpt: > the lexemes are just not going to implement it, and that's fine (they will not be usable in the contexts where this interface is required).
Yes, that's true.

I still don't like having such interface. But let's introduce it now and I'll try to figure out a way to drop it.
I also believe that we will want to get free of the hugly StatementUpdate and TermedStatementUpdate classes as soon as possible.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Statement.java
Tpt: > I think we rely on this distinction in StatementUpdate (but I am not entirely sure)

The Wikibase API returns the statement serialization when created or modified. I believe that in StatementUpdate what would be clean to do is just deserialize the returned statements and replace the pushed statement by the new one. This way we would avoid this setter and also get the value
 normalizations routines executed by wikibase also pushed back in the Java object.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
wetneb: yeah I want to refactor `StatementUpdate` and `TermedStatementUpdate` too, and introduce custom matching strategies for statements, qualifiers and references (the current matching mode is not always adequate)

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/EntityDocument.java
Tpt: Ok, I see.

> One thing we could do is merge the two methods: each with method of EntityDocument, StatementDocument, etc would also take the new revision id as a second parameter. 

It looks quite heavy but why not.

An other question: do we want to keep the revision id if the entity is modified without new revision id given? If we remove it it would have the advantage of not having entity objects in a state that is not the one of their revision id. But it would require the toolkit user to give the base revision id when trying to save an edited entity.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
Tpt: Great! So, let's keep this interface for now

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/EntityDocument.java
wetneb: so you would silently "reset" the revision id in `withDescription` for instance? users might not expect that…

if we go for `withDescription(desc, newRevId)` then I believe we can get rid of `TermedStatementDocument` altogether (because there will be no intersection between the methods of `StatementDocument` and `TermedDocument`).

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Statement.java
wetneb: no no, we actually need to set the ID (generated by us) *before* sending the statement to Wikibase. That's really weird, but that's how the API works. They expect clients to generate these ids.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
wetneb: by the way what sort of structure are you thinking about for `StatementUpdate`? Beyond adding a builder for it?

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Statement.java
Tpt: Thank you. I have forgotten that you had to create a statement id when you want to create a statement with qualifier and references in one edit. Ok, so it make sense to have this method.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
Tpt: I was thinking about this workflow:

Item item = ...
item = item.withLabel("en", "Foo");
item = entityUpdater.saveItem(item, "my edit comment")

entityUpdater would do the diff with the item content in Wikidata for the relevant revisionId and pick the right edit action for you (and fail if there is an edit conflict).
But I am not sure that it is the best workflow. Have you an other in mind?

wetneb: hmm ok, so that `Item` class would keep track of both the initial `ItemDocument` and the new one. That would be very nice to use indeed! But AFAICT it's just about adding a new interface on top of the existing system (StatementUpdate would just not be manipulated by the user directly).

the refactoring I want to introduce is just to let users specify matching modes with each statement that they want to add. For instance, they might want to say "add this statement but only merge the reference with any existing statement if that existing statement has no reference yet", or "add this statement by matching only on mainsnak and merging qualifiers"… things like that.

Tpt: > hmm ok, so that Item class would keep track of both the initial ItemDocument and the new one. 

I would avoid that. It's too much responsibilities for this interface.

> the refactoring I want to introduce is just to let users specify matching modes with each statement that they want to add. For instance, they might want to say "add this statement but only merge the reference with any existing statement if that existing statement has no reference yet", or "add this statement by matching only on mainsnak and merging qualifiers"… things like that.

It looks great! We should probably move into this direction instead.

wetneb: > I would avoid that. It's too much responsibilities for this interface.

Yeah I did not mean putting both of these things in `ItemDocument` (I thought you had introduced a distinction between `Item` and `ItemDocument`).

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
```diff
@@ -0,0 +1,53 @@
...
+	TermedStatementDocument withRevisionId(long newRevisionId);
+	
+	@Override
+	TermedStatementDocument withLabel(MonolingualTextValue newLabel);
+	
+	@Override
+	TermedStatementDocument withDescription(MonolingualTextValue newDescription);
+	
+	@Override
+	TermedStatementDocument withAliases(String language, List<MonolingualTextValue> aliases);
```
Tpt: Is there a good reason to have the language code as first argument even if it is already contained in `MonolingualTextValue`?
What about just a simple `TermedStatementDocument withAlias(MonolingualTextValue alias);` method?

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/StatementDocument.java
```diff
@@ -479,4 +479,33 @@ MonolingualTextValue findStatementMonolingualTextValue(
...
+	
+	/**
+	 * Returns a new version of this document where all statements matching
+	 * any of the statement ids provided have been removed. These statements
+	 * can use different properties.
+	 * 
+	 * @param statementIds
+	 *       the identifiers of the statements to remove
+	 */
+	StatementDocument withoutStatementIds(Set<String> statementIds);
```
Tpt: What about just having `StatementDocument withoutStatement(String statementId)`.  The interface would be simpler (but the execution when multiple statements are removed a bit less efficient indeed even if it's probably negligible against the cost of pushing edits to Wikidata).

wetneb: Sure!

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedStatementDocument.java
wetneb: The thing is that this *replaces* all aliases in the given language by the list supplied. This gives more control, because it lets people remove all aliases by passing the empty list (hence the language parameter is needed).

Tpt: I see. What about having more explicit methods like `withoutAliases(String language)` that removes all aliases in a language and `withAlias(MonolingualTextValue alias)` to add an alias (and maybe `withoutAlias(MonolingualTextValue alias)` to remove an alias?

wetneb: With these methods you still cannot control the order of the aliases, which is stored by wikibase…

Tpt: Good point. But it is not significant on the semantic level so I hope nobody cares. But I'm ok with your current method if you are afraid of this problem.

wetneb:
Cool, I'll clean up for the remaining issues you raised and do that.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/StatementGroupImpl.java
```diff
@@ -107,4 +109,23 @@ public String toString() {
 		return ToString.toString(this);
 	}
 
+	@Override
+	public StatementGroup withStatement(Statement statement) {
+		String statementId = statement.getStatementId();
```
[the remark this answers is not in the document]
wetneb: Isn't this already checked by the constructor that we call later on? I think these checks should only be in the constructor (so, we really rely on immutability).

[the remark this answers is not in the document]
Tpt: Indeed. My bad.

Tpt:
Hey! Thanks for the rebase! It seems there is still a merge conflict :-(

PR #326, opened 2018-03-03
Cleanup a few issues with JSON serialization after the refactoring

wetneb:
I have noticed a few issues with the refactoring of the datamodel:
- qualifiers were not serialized correctly anymore. I have preventively added explicit `@JsonProperty` on the getters where it was missing to prevent that from happening at other places (and added a test `testFullStatementToJson` for that)
- `SnakImpl` was deconstructing and reconstructing its `PropertyIdValue` for no reason, this is now fixed

As a result another bug resurfaced: `ItemIdValue.NULL` and `PropertyIdValue.NULL` were not hashed properly (that's independent from the rewrite - although I have to admit that I am responsible for adding `equals` there, without which the bug would not appear either… but surely we do want to have both `equals` and `hashCode` there, right?).


Tpt:
>  but surely we do want to have both equals and hashCode there, right?).
Yes, definitely

PR #329, opened 2018-03-06
Adds Claims methods to Statement

Tpt:
Makes easier for clients to retrieve statements content without requiring an unneeded getter.

Use JaksonPreStatement only for deserialization

Drops some broken tests of DatamodelConverter: the input validation is now stronger


Tpt:
Needs extra work

PR #330, opened 2018-03-06
Adds Claims methods to Statement

Tpt:
Makes easier for clients to retrieve statements content without requiring an unneeded getter.

Uses JaksonPreStatement only for deserialization

### DIFF
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/json/JacksonPreStatement.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/json/JacksonPreStatement.java
@@ -21,10 +21,7 @@ package org.wikidata.wdtk.datamodel.implementation.json;
  */
 
 import com.fasterxml.jackson.annotation.*;
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import org.apache.commons.lang3.Validate;
 import org.wikidata.wdtk.datamodel.implementation.ReferenceImpl;
 import org.wikidata.wdtk.datamodel.implementation.SnakImpl;
 import org.wikidata.wdtk.datamodel.implementation.StatementImpl;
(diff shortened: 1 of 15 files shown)


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/json/JacksonPreStatement.java
```diff
@@ -33,100 +30,53 @@
...
-		if (propertyOrder != null) {
-			this.propertyOrder = propertyOrder;
-		} else {
-			this.propertyOrder = Collections.emptyList();
-		}
-		if (references != null) {
-			this.references = references;
-		} else {
-			this.references = Collections.emptyList();
-		}
```
wetneb: Are we still be able to deserialize statements where references or qualifiers are not provided? (Maybe nobody produces JSON like that actually?)

Tpt: Yes, we are still able to deserialize these statements (See the implementation.json.TestStatement tests). It does not cost us much to still support this case and there is a meaningful fallback (no references/qualifiers)

wetneb: okay so Jackson automatically passes empty containers for properties that are not supplied?

Tpt: It passes null for the not supplied properties. It's why we need to check for null.

wetneb: That's what I thought - so where are these checks gone?

Tpt: In the StatementImpl constructor. No need to have them twice.

wetneb: sure, thanks

Tpt:
> This shortcut going to save me a lot of typing in the years to come! :-D

Me too ;)

> Why was Claim introduced in the first place btw?

Two reasons:
1. It was planned that for some use cases that would not require support for multiple point of views plain claims would be used instead of statements. But the Wikidata dev team made the choice to remove this option in the Wikibase implementation a few years ago.
2. It makes sense to define this object because references are sources for a claim, not just of a snak or something else. It makes the description of the data model easier.

PR #338, opened 2018-03-11
[Breaking] Uses "1" as the identifier for empty unit

Tpt:
Consistent with the PHP implementation: https://github.com/DataValues/Number/blob/master/src/DataValues/UnboundedQuantityValue.php#L41

It has the advantage of removing one of the easy to miss difference between WikidataToolkit and the API output but the disadvantage of being a braking change compared to the released versions.


wetneb:
@Tpt would it be worth mentioning that in the changelog?

Tpt:
yes, definitely, alongside many other things. I was planning to update the changelog before releasing the next version using the commit history (there are many more smaller braking changes that also appends).

PR #348, opened 2018-04-03
Drops WikibaseClient

Tpt:
Not useful anymore


Tpt:
@mkroetzsch @guenthermi Do you still need this client for SQID? imho this code should be in a project that relies on WikidataToolkit and not in WikidataToolkit itself

mkroetzsch:
Yes, SQID is currently using the client to extract class and property information from the dumps.

mmarx:
For what it's worth, I have forked the code and am still planning on splitting out the client into a dedicated SQID helper; it needed a fix anyway to comply with the user agent policy. In the meantime, do you still want a pull request for that fix?

Tpt:
@mmarx If you have forked the code then I believe we could safely drop this package from WikidataToolkit. I am not sure that the conversion code to JSON/RDF dump is useful anymore to anyone now that Wikidata provides JSON and RDF dumps.

mkroetzsch:
@tpt Wikidata only provides huge RDF dumps that include all data. A student of mine is currently working on a web service to generate smaller, custom RDF dumps (as needed for many research papers and applications), and this work is based on WDTK's RDF generation. There might also be some pull requests towards updating this part of WDTK for better alignment with the current Wikidata export.

Tpt:
@mkroetzsch I was not planning to drop the wdtk-rdf package. I rely also on it for http://wdhqs.wmflabs.org/

PR to update the RDF converter are much welcome!

mmarx:
Okay then, feel free to reopen this at your leisure. :)

wetneb:
If wdtk-client can be dropped safely, I would be keen to do that (for instance for the coverage "boost" reported by Coveralls above!).

mmarx:
The client code is now part of the SQID repository.

Tpt:
@mmarx Thanks! Let's remove WikibaseClient

PR #353, opened 2018-04-09
Makes *Impl constructors care about TermImpl vs MonolingualTextValueImpl

Tpt:
There are two different implementations of `MonolingualTextValue`, one for snaks and one for labels/description/aliases. This PR makes sure that the constructors cares that their parameters are using the right implementation.


Tpt:
I still believe that terms and monolingual text values are still the same kind of thing (i.e. language-tagged strings), even if, indeed, they play different roles. But it seems to me that there is as much difference between a term used as label and a term used as description as between a term used as description and a monolingual text value used e.g. for a "pronunciation" property.

Tpt:
The abstract description of the Wikibase DataModel is also not doing any distinction between Term and MonolingualTextValue: https://www.mediawiki.org/wiki/Wikibase/DataModel

PR #355, opened 2018-04-12
Moves out and reprecate DatamodelConverter

Tpt:
Now that we have only one Datamodel implementation, the DatamodelConverter classes looses a part of its goals. This PR:
* adds a new class DatamodelFilter to implement the filtering features of DatamodelConverter
* drops or migrates all DatamodelConverter usages in the code
* provides okish tests of DataObjectFactoryImpl without relying on DatamodelConverter
* deprecates DatamodelConverter


Tpt:
The code coverage decreased because I removed a lot of `assertEquals(o1.toString(), o2.toString()) `where `o1.equals(o2)`. I do not think it is a bad loss

wetneb:
I am still using the DatamodelConverter (it is useful when you define your own implementations of various classes of the datamodel and want to convert them later on to the default implementation), but I can easily copy that to OpenRefine.

PR #371, opened 2018-05-15
Makes clearer the cases where Snak.getValue returns null 

Tpt:
Avoids a getter that may return null

### DIFF
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Snak.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/Snak.java
@@ -45,7 +45,7 @@ public interface Snak {
 	 * Get the {@link Value} of this Snak, or null if the snak has no specified
 	 * value.
 	 *
-	 * @return Value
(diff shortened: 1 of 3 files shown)


wetneb:
How about converting it to use Optionals instead? I find it quite convenient to have this method in Snak, otherwise it's a bit annoying to have to do a cast.

Tpt:
> How about converting it to use Optionals instead?

It's a good idea but it would be a breaking change. Having this method is indeed useful. I'm going to modify this pull request to remove the deprecation but keep having an override in ValueSnak that states that the method never returns null.

wetneb:
You're right about the breaking change… it's going to be annoying… then I don't know. I guess we can deprecate but when we do so we need to provide a viable alternative at the same time. So I would be happy with deprecating this one and adding another method (such as `maybeGetValue` or something better) with Optionals.

Tpt:
>  So I would be happy with  deprecating this one and adding another method (such as maybeGetValue or something better) with Optionals.

It sounds like a good option. We would keep the getValue method not deprecated for the ValueSnak interface that always have a value.

Tpt:
Let's do an other pull request for the optional getter

PR #373, opened 2018-05-15
Adds utility method to get unit, globe and preferred calendar as ItemIdValue

Tpt:
I am not sure how to properly implement and expose the entity IRI
parsing so the first implementation is kept in a private place


Tpt:
>  On the long term I would ideally like that the default implementations directly store these things as ItemIdValue
It's probably not a good idea because it could be any string and there is sometime garbage in this field. So, we would need to have a sensible fallback in these cases instead of just failing dump parsing.

PR #378, opened 2018-06-19
Implements lexemes senses

Tpt:


wetneb:
whoops I merged this without realizing that it broke compilation - it is now fixed, sorry about that!

PR #382, opened 2018-07-04
Deprecates Snak.getValue() in favor of ValueSnak.getValue()

Tpt:
Avoids a getter that may return null


Tpt:
It would be maby indeed nice to have such alternative but Optional is hardly available on Android and doing `if(snak instanceof ValueSnak)` instead of `if(snak.getValue() == null)` does not seeems very bad to me.

wetneb:
Good to know!

PR #384, opened 2018-07-04
Use same kind of values IRIs as the official Wikidata RDF format

Tpt:
The value hash used is still not the same


Tpt:
@wetneb It would be nice to generate the exact same URIs as Wikibase but the way Wikibase generate them is actually doing md5(serialize($myPhpObject)) which is hardly portable to other languages. I merge this PR because it is already a step in the good direction.

wetneb:
arf - sad that it is language dependent indeed

Issue #268, opened 2017-07-26
OAuth support


wetneb:
Hi,

First, this library looks fantastic! So much cleaner than the ones we have in Python. Thank you for that!

Any chances logging in with OAuth could be supported? That would be massively useful. As far as I can tell ApiConnection only works with standard credentials for now.

Keep up the good work!

Tpt:
> First, this library looks fantastic! So much cleaner than the ones we have in Python. Thank you for that!

Thank you!

> Any chances logging in with OAuth could be supported?

We should definitely support OAuth. Feel free to write a pull request.

wetneb:
@Tpt great! Not sure my Java is fluent enough to deserve to be included in that library, and pretty sure it would be very unreasonable to procrastinate my PhD on that, but I'll definitely consider it!

lucaswerkmeister:
scribejava/scribejava#852 adds a class describing MediaWiki’s OAuth URLs to the ScribeJava API. Since it might be a while until that pull request is merged and the next version of the library is released, I hereby as the original and sole author of the [`MediaWikiApi` class](https://github.com/lucaswerkmeister/scribejava/blob/mediawiki/scribejava-apis/src/main/java/com/github/scribejava/apis/MediaWikiApi.java) grant you permission to copy that class into Wikidata Toolkit (ideally with a different package name) and start using it right now :)

(Note: the pull request also adds a `MediaWikiExample` class, but since that is mostly derived from other examples in the repository, I cannot grant you a license for that.)

lucaswerkmeister:
If I was allowed to make breaking changes, I think I would implement it a bit like this (sublists indicate subclasses):

- `ApiConnection`: abstract class with one abstract method (`sendJsonRequest`) and a few static utility methods (`implodeObjects`, and perhaps the two current wikidata/testwikidata getters?)
  - `BasicApiConnection`: most of the current `ApiConnection` code – implements `sendJsonRequest` and manages cookies, but doesn’t do login (can be used for anonymous requests)
    - `UsernamePasswordApiConnection`: adds login/logout to `BasicApiConnection` by sending API requests which populate `BasicApiConnection`’s cookie store (the rest of the current `ApiConnection` code)
  - `OAuthApiConnection` – has methods for the three-legged OAuth flow using ScribeJava, and hopefully using some nice HTTP library (ScribeJava seems to come with something by default?) so we don’t need to bother with cookies etc. ourselves

[...]

PR #410, opened 2019-03-05
Throw an exception when credentials have expired.

### LINKED ISSUE #409: Editing continues as an IP if the auth cookies expire
If we login with WDTK and then the cookies expire (for instance because we logout from the same account with a browser), then editing in WDTK will continue as an IP, which is not desirable.

The `assert=` parameter can be used to make sure the action fails if we are not authenticated as the given user. We should use that:
https://www.mediawiki.org/wiki/API:Assert

wetneb:
See https://www.mediawiki.org/wiki/API:Assert.
Fixes #409.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -184,7 +184,6 @@ public class ApiConnection {
 	 */
 	@JsonIgnore
 	String password = "";
-	
 	/**
 	 * Map of cookies that are currently set.
 	 */
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -161,6 +161,11 @@ public class ApiConnection {
 	final static String LOGIN_WRONG_TOKEN = "WrongToken";
 
 	/**
+	 * MediaWiki assert= parameter to ensure we are editting while logged in.
+	 */
+	private static final String ASSERT_PARAMETER = "assert";
+
+	/**
 	 * URL to access the Wikibase API.
 	 */
 	final String apiBaseUrl;
@@ -179,7 +184,7 @@ public class ApiConnection {
 	 */
 	@JsonIgnore
 	String password = "";
-
+	
 	/**
 	 * Map of cookies that are currently set.
 	 */
@@ -394,6 +399,9 @@ public class ApiConnection {
 	 */
 	public JsonNode sendJsonRequest(String requestMethod, Map<String,String> parameters) throws IOException, MediaWikiApiErrorException {
 		parameters.put(ApiConnection.PARAM_FORMAT, "json");
+		if (loggedIn) {
+			parameters.put(ApiConnection.ASSERT_PARAMETER, "user");
+		}
 		try (InputStream response = sendRequest(requestMethod, parameters)) {
 			JsonNode root = this.mapper.readTree(response);
 			this.checkErrors(root);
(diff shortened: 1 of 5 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/apierrors/AssertUserFailedException.java
```diff
@@ -0,0 +1,37 @@
...
+ * credentials have expired. See
+ * <a href="https://www.mediawiki.org/wiki/API:Assert">MediaWiki documentation</a>.
+ * 
+ * @author Antonin Delpeuch
+ *
+ */
+public class AssertUserFailedException extends MediaWikiApiErrorException {
+
+	public AssertUserFailedException(String errorMessage) {
+		super(MediaWikiApiErrorHandler.ERROR_NO_SUCH_ENTITY, errorMessage);
```
Tpt: Wrong error constant
```suggestion
		super(MediaWikiApiErrorHandler.ERROR_ASSERT_USER_FAILED, errorMessage);
```

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -179,7 +184,7 @@
 	 */
 	@JsonIgnore
 	String password = "";
-
+
```
Tpt: training whitespace
```suggestion

```

wetneb:
@Tpt yes I was wondering about that too! One the one hand I think it is nice that the `assert` is added by the `ApiConnection` and not the `WbEditingAction`, as this really is an authentication concern. On the other hand it does apply the parameter for all requests, even the non-editing one. That being said, in general, you do want to make sure that you are logged in for non-editing actions too: for instance, assuming you have a bot flag, this will let you request batches of 500 entities instead of 50 with `wbgetentities`.

Maybe `ApiConnection` should expose two methods (or one with a boolean parameter) to tell if we want to make the given request as a logged-in user or not. Dynamically determining that based on the query parameters inside `ApiConnection` sounds sloppy to me, there are so many different API actions that we are likely to miss some corner cases.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/apierrors/MediaWikiApiErrorHandler.java
```diff
@@ -33,6 +33,7 @@
 	public final static String ERROR_INVALID_TOKEN = "badtoken";
 	public final static String ERROR_NO_SUCH_ENTITY = "no-such-entity";
 	public final static String ERROR_MAXLAG = "maxlag";
+	private static final String ERROR_ASSERT_USER_FAILED = "assertuserfailed";
```
Tpt: It should be package private to be able to reuse it in AssertUserFailedException definition.

Tpt:
Thank you for your comment. I believe it make sense to add the verification to all queries to keep the API simple and avoid too many options. What do you think about it?

wetneb:
Sure, that sounds reasonable to me. If users have special needs it should always be possible to hold two `ApiConnection` objects, one logged in and the other logged out.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/apierrors/MediaWikiApiErrorHandler.java
wetneb: Oops, sorry! I have made it public to match the rest of the codes above.

Tpt:
Thank you! Do you want to get a release quickly for inclusion in OpenRefine?

wetneb:
That would be helpful, thanks. I thought maybe I could squeeze in #268 too?

Tpt:
> I thought maybe I could squeeze in #268 too?

It would be great!

PR #418, opened 2019-06-14
Fix Statement serialization.

### LINKED ISSUE #417: Regression: serialization of Statements
We have a problem with Statement serialization: statement values should only be serialized as part of the mainsnak, but they are also serialized at the Statement root:

Our tests currently specify this serialization:
```json
{
  "rank": "preferred",
  "value": {
    "value": {
      "id": "Q42",
      "numeric-id": 42,
      "entity-type": "item"
    },
    "type": "wikibase-entityid"
  },
  "id": "MyId",
  "mainsnak": {
    "property": "P42",
    "datatype": "wikibase-item",
    "datavalue": {
      "value": {
        "id": "Q42",
        "numeric-id": 42,
        "entity-type": "item"
      },
      "type": "wikibase-entityid"
    },
    "snaktype": "value"
  },
  "type": "statement"
}
```

The correct serialization should be:
```json
{
  "rank": "preferred",
  "id": "MyId",
  "mainsnak": {
    "property": "P42",
    "datatype": "wikibase-item",
    "datavalue": {
      "value": {
        "id": "Q42",
        "numeric-id": 42,
        "entity-type": "item"
      },
      "type": "wikibase-entityid"
    },
    "snaktype": "value"
  },
  "type": "statement"
}
```

This was introduced by the addition of Claim methods to the Statement class, #330, where a `@JsonIgnore` was forgotten. The faulty test was introduced in 5d34b3a0c.

wetneb:
This removes the duplicate statement value serialization.

Closes #417.


wetneb:
By the way I find it quite surprising that this does not affect other tests. Our test suite could do with a bit more redundancy to make it more robust, I would say. Messing up with the JSON serialization should make Travis really really angry.

All our editing tests mock their requests by generating the expected JSON themselves with Jackson. That JSON should be frozen in files, so that the tests really specify the API we are working against.
We should have tests serializing and deserializing real-life ItemDocuments in their entirety. That would catch bugs like this one and a lot of others.

So, note to self:
* We need a lot more integration tests!
* Our test suite is not great, do not trust it.

Tpt:
Indeed. We should definitely improve the test suite. Thank you for spotting the bug!

Issue #407, opened 2018-11-28
Nullpointer Exception - RDF Serialization


Livvi:
Hey,

I am trying to serialize RDF Tripels of the 2018-11-05 json dump, but it always stops after getting this NullPointerException: 

```
WARN  - Could not determine Type of P5193. It is not a valid EntityDocument Id
Exception in thread "main" java.lang.NullPointerException
	at org.wikidata.wdtk.rdf.values.EntityIdValueConverter.getRdfValue(EntityIdValueConverter.java:46)
	at org.wikidata.wdtk.rdf.values.AnyValueConverter.visit(AnyValueConverter.java:95)
	at org.wikidata.wdtk.rdf.values.AnyValueConverter.visit(AnyValueConverter.java:49)
	at org.wikidata.wdtk.datamodel.implementation.FormIdValueImpl.accept(FormIdValueImpl.java:125)
	at org.wikidata.wdtk.rdf.values.AnyValueConverter.getRdfValue(AnyValueConverter.java:90)
	at org.wikidata.wdtk.rdf.SnakRdfConverter.visit(SnakRdfConverter.java:150)
	at org.wikidata.wdtk.rdf.SnakRdfConverter.visit(SnakRdfConverter.java:53)
	at org.wikidata.wdtk.datamodel.implementation.ValueSnakImpl.accept(ValueSnakImpl.java:138)
	at org.wikidata.wdtk.rdf.RdfConverter.writeClaim(RdfConverter.java:391)
	at org.wikidata.wdtk.rdf.RdfConverter.writeStatement(RdfConverter.java:369)
	at org.wikidata.wdtk.rdf.RdfConverter.writeStatements(RdfConverter.java:282)
	at org.wikidata.wdtk.rdf.RdfConverter.writePropertyDocument(RdfConverter.java:190)
	at org.wikidata.wdtk.rdf.RdfSerializer.processPropertyDocument(RdfSerializer.java:141)
	at org.wikidata.wdtk.datamodel.interfaces.EntityDocumentProcessorBroker.processPropertyDocument(EntityDocumentProcessorBroker.java:59)
	at org.wikidata.wdtk.dumpfiles.JsonDumpFileProcessor.handleDocument(JsonDumpFileProcessor.java:129)
	at org.wikidata.wdtk.dumpfiles.JsonDumpFileProcessor.processDumpFileContents(JsonDumpFileProcessor.java:88)
[...]

Issue #422, opened 2019-08-01
Support for unknown datavalues


wetneb:
Support for new datatypes is regularly added to Wikidata.

Every time this happens, previous versions of WDTK will be unable to deserialize EntityDocuments where these datatypes are used. This means that code written using WDTK will choke on these entities, even if the particular task at hand does not do anything with these statements with unknown datatypes. This seems less than ideal to me.

In principle, WDTK should be able to deserialize datavalues with unknown datatypes by just by recording their JSON representation, so that these values can be serialized back in the same format. The Java representation of these datavalues would be opaque to the user - they would not be able to manipulate them in any useful way. So as a user these datatypes would of course still be "unsupported" but their presence in an EntityDocument would not prevent you from doing work that does not interact with the values.

That design should work fine for items, but it would still choke on `PropertyDocument`s with unknown datatypes. That is less of a concern as far as I can tell, but ideally it would be nice to handle this case as well.

@Tpt any thoughts?

wetneb:
Actually `DatatypeIdImpl` seems to handle unknown datatypes gracefully at the moment, so it is just a matter of adding a new `ValueImpl` subclass, with the appropriate visibility.

wetneb:
The datatypes recently introduced by WMDE are:
- lexeme, forms and senses -> these are new entity ids, causing issues mentioned in #407 on old versions of WDTK
- musical notation, mathematical formula, and other similar datatypes, which use strings as underlying datavalue -> WDTK did not choke on these.

However, I think it would be useful to have a fallback for all possible datatypes, not just for new entity ids of new types.

Tpt:
I believe we need to keep in mind the distinction that Wikibase does between datavalues and datatypes.

1. DataValues are the data structures. Currently the ones used by Wikidata are "string", "monolingual-text", "wikibase-entityid", "time", "globe-coordinates" and "quantity". They are presented in WikidataToolkit by a sub interface of `Value` and WikidataToolkit hard fails on unknown datavalues.

2. DataTypes that are a "type" "tag" stored in the property descriptions and returned as part of the snak serialization by some Wikibase API/dumps. For example, "wikibase-item", "wikibase-property"... are possible datatypes for the "wikibase-entityid" data value type, "commons-media" or "external-id" for "string"... They are presented by the `DatatypeId` interface and WikidataToolkit should not break if a new datatype is introduced.

[...]

PR #423, opened 2019-08-01
Support unsupported values (!)

wetneb:
This adds the fallback classes discussed in #422. I chose the word `Unsupported` rather than `Unknown` since I felt that the latter could be mistaken with the notion of SomeValue/NoValue snaks (users should not try to create a SomeValue snak by instantiating these fallback classes and using it as snak value…)

This also fixes issue #407, which was actually a genuine issue. Postmortem for that one:
- The warning message in the `wdtk-rdf` module was really not informative at all, so it made it harder to debug
- The code that infers a datatype from an entity value had not been updated when lexeme support was added - but really that code should not live in the `wdtk-rdf` module, it duplicates things we already have (but do not expose) in the `wdtk-datamodel` module. That code is now updated but we should really consider removing this duplication by exposing this ugly (but useful) heuristic in the datamodel module.

### DIFF
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
@@ -23,6 +23,7 @@ package org.wikidata.wdtk.datamodel.implementation;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.lang3.Validate;
 import org.wikidata.wdtk.datamodel.helpers.Equality;
 import org.wikidata.wdtk.datamodel.helpers.Hash;
 import org.wikidata.wdtk.datamodel.helpers.ToString;
@@ -34,6 +35,8 @@ import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.JsonDeserializer.None;
@@ -77,12 +80,14 @@ public class UnsupportedEntityIdValueImpl extends ValueImpl implements Unsupport
 				String entityType,
 				@JsonProperty("id")
 				String id) {
+			Validate.notNull(id);
 			this.entityType = entityType;
 			this.id = id;
 			contents = new HashMap<>();
 		}
 		
 		@JsonProperty("entity-type")
+		@JsonInclude(Include.NON_NULL)
 		public String getEntityTypeString() {
 			return entityType;
 		}
@@ -154,7 +159,7 @@ public class UnsupportedEntityIdValueImpl extends ValueImpl implements Unsupport
 
 	@Override
 	@JsonIgnore
-	public String getEntityTypeString() {
+	public String getEntityTypeJsonString() {
 		return value.getEntityTypeString();
 	}
 }
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
@@ -1,5 +1,25 @@
 package org.wikidata.wdtk.datamodel.implementation;
 
+/*-
+ * #%L
+ * Wikidata Toolkit Data Model
+ * %%
+ * Copyright (C) 2014 - 2019 Wikidata Toolkit Developers
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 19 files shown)


wetneb:
Also I should mention that my fix of #407 serializes unknown values with blank nodes - the assumption being that we do not want to stop an entire RDF export just because some rare values are not supported. We can also decide to stop the export altogether by throwing an exception.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
```diff
@@ -0,0 +1,160 @@
...
+
+	private static class JacksonIdValue {
+		private final String entityType;
+		private final String id;
+		private final Map<String, JsonNode> contents;
+		
+		@JsonCreator
+		private JacksonIdValue(
+				@JsonProperty("entity-type")
+				String entityType,
```
Tpt: I believe that now "id" is the only key that is always supposed to exist. We should make sure we do not fail in this case.

wdtk-datamodel/src/test/java/org/wikidata/wdtk/datamodel/implementation/ItemIdValueImplTest.java
```diff
@@ -131,10 +136,11 @@ public void testToJavaWrongID() throws IOException {
 		mapper.readValue(JSON_ITEM_ID_VALUE_WRONG_ID, ValueImpl.class);
```
Tpt: I believe that we want this test to return a UnsupportedEntityIdValue too

wdtk-datamodel/src/test/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueTest.java
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class UnsupportedEntityIdValueTest {
+	private final ObjectMapper mapper = new DatamodelMapper("http://www.wikidata.org/entity/");
+
+	private final String JSON_UNSUPPORTED_VALUE_1 = "{\"type\":\"wikibase-entityid\",\"value\":{\"entity-type\":\"funky\",\"id\":\"Z343\"}}";
+	private final String JSON_UNSUPPORTED_VALUE_2 = "{\"type\":\"wikibase-entityid\",\"value\":{\"entity-type\":\"shiny\",\"id\":\"P8989\",\"foo\":\"bar\"}}";
+
```
Tpt: Would be nice to have a test with only "id"

wdtk-datamodel/src/test/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedValueImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+public class UnsupportedValueImplTest {
+	private final ObjectMapper mapper = new ObjectMapper();
+
+	private final String JSON_UNSUPPORTED_VALUE_1 = "{\"type\":\"funky\",\"value\":\"groovy\"}";
+	private final String JSON_UNSUPPORTED_VALUE_2 = "{\"type\":\"shiny\",\"number\":42}";
+	
+	private UnsupportedValue firstValue, secondValue;
+	
+	@Before
+	public void deserializeFirstValue() throws JsonParseException, JsonMappingException, IOException {
```
Tpt: This method name seems a bit misleading.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
```diff
@@ -0,0 +1,160 @@
...
+	
+	@Override
+	public String toString() {
+		return ToString.toString(this);
+	}
+
+	@Override
+	@JsonIgnore
+	public String getEntityType() {
+		return ET_UNSUPPORTED;
```
Tpt: It might be relevant to return null or something build from entityType

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
wetneb: Makes sense. Do you know in what scenario the `entity-type` can go missing?

Tpt: I believed it was the case for of Lexemes, but it seems it's not the case anymore. Example: https://www.wikidata.org/wiki/Special:EntityData/L2.json

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
wetneb: Yeah, I was not too sure. In general I am not a big fan of making up recipes to build these sort of URIs, unless the rules for that are documented somewhere? Returning `null` could also cause NPEs potentially…

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/UnsupportedValue.java
```diff
@@ -0,0 +1,43 @@
...
+public interface UnsupportedValue extends Value {
+	
+	/**
+	 * Returns the type string found in the JSON representation
+	 * of this value.
+	 * 
+	 * @return
+	 * 		the value of "type" in the JSON representation of this value.
+	 */
+	public String getTypeString();
```
Tpt: Maybe `getTypeJsonString` to be consistent with `UnsupportedEntityValue`? Or keep getTypeString in `UnsupportedEntityValue`

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/UnsupportedEntityIdValueImpl.java
Tpt: Ok! Make sense!

Tpt: Sorry to change my mind. Entity types URIs are always following the pattern `"http://wikiba.se/ontology#Wikibase" + stirngEntityType.toCamlCase()`. It is exactly the same conversion that is implemented in `DatatypeIdImpl`.

wetneb: Fine with me!

PR #425, opened 2019-08-02
Fix coveralls upload on Travis, use SVG badge.

wetneb:
Coverage upload failed because of https://github.com/trautonen/coveralls-maven-plugin/issues/127.
See failing builds: https://travis-ci.org/Wikidata/Wikidata-Toolkit/jobs/566845341#L1823

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,11 @@ jdk:
   - oraclejdk9
   - oraclejdk11
 
+script:
+  - mvn cobertura:cobertura
+
 after_success:
-  - mvn clean cobertura:cobertura coveralls:cobertura
+  - mvn coveralls:cobertura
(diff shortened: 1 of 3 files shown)


wetneb:
Hmm actually downgrading the Maven plugin version to 2.7 makes Travis execute the tests twice, it's not ideal. Let's try to execute them only once.

wetneb:
Actually cobertura-maven-plugin does not work for recent versions of Java: https://github.com/mojohaus/cobertura-maven-plugin/issues/43. We need to switch to something else.

thadguidry:
Yeah we switched to Jacoco almost exclusively at work, especially since it was so painful to get Gradle to work well with Cobertura.

Issue #424, opened 2019-08-02
Code duplication: determine an entity type by looking at the format of its id


wetneb:
There are at least two places in our code where we try to guess the type of an entity based on the format of its id:

https://github.com/Wikidata/Wikidata-Toolkit/blob/ac2692ac35e3ae7fc8148f416322e1e6f07beca0/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java#L134-L162

and

https://github.com/Wikidata/Wikidata-Toolkit/blob/d76e8ee421856f7aba6bff6cf1f8ab13e03f451d/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/PropertyRegister.java#L199-L229

That is a problem: it is easy to forget to update one of them (it caused #407).
We should expose this functionality publicly in the `wdtk-datamodel` module. This raises questions:
* where should this static method live?
* what representation of the entity types should it return? Currently one uses the JSON strings, the other uses IRIs for datatypes. I think it should be IRIs for entity types since that is what the `EntityIdValue` interface offers… But then we probably need other static methods to help convert between all these formats? It's a bit of a mess :-/

Tpt:
Yes, indeed it's a bit of a mess. The choice have been done at the beggining of WikidataToolkit to create URIs for each Wikibase datamodel enumeration (datatypes, entity types...), leading to this duplication. We have two options here:

1. Drop everywhere the URIs and only returns the JSON string ids that are used by Wikibase. It has the strong advantage of not having to update the mappings each time a new value is introduced.
2. Keep the library working as it is.

I would slightly prefer 2. because it is less disruptive for the library user. In this case we should probably have a method that returns the Datatype IRI. This method could be public in the `EntityIdValue` interface. We could then have package private conversion methods between the two representation in EntityIdValueImpl, just like it is done in `DatatypeIdImpl` for the datatype ids.

wetneb:
Option 2 is fine with me too.
Since the entity type IRIs are also exposed publicly in `DatatypeIdValue` I would say that it would be good to have a public conversion utility, transforming `http://wikiba.se/ontology#WikibaseForm` (datatype IRI) to `http://www.wikidata.org/ontology#Form` (entity type IRI) and back. Conversion to JSON strings can stay private since they are not exposed except in the pathological case of unsupported entities.

snehasi:
Hi

[...]

PR #427, opened 2019-08-06
 Adds MediaInfoValueId and MediaInfoDocument

Tpt:


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/Datamodel.java
```diff
@@ -183,6 +188,34 @@ public static SenseIdValue makeWikidataSenseIdValue(String id) {
...
+	 *            IRI of the site this belongs to, e.g.,
+	 *            "http://www.wikidata.org/entity/"
+	 * @return an {@link MediaInfoIdValue} corresponding to the input
+	 */
+	public static MediaInfoIdValue makeMediaInfoIdValue(String id, String siteIri) {
+		return factory.getMediaInfoIdValue(id, siteIri);
+	}
+
+	/**
+	 * Creates an {@link LexemeIdValue} for Wikimedia Commons.
```
wetneb: LexemeIdValue should be MediaInfoIdValue

wetneb:
Awesome! Thanks a lot for this!!

Potentially it could be worth to update the wdtk-rdf module (see #424) for datatype detection although it is not likely to be useful immediately (I don't think mediainfo are meant to be used as statement values? - although you do add support for that sort of deserialization in `MediaInfoIdValueImpl`, which cannot hurt).

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/MediaInfoDocumentImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Constructor. Creates an object that can be populated during JSON
+	 * deserialization. Should only be used by Jackson for this very purpose.
+	 */
+	@JsonCreator
+	public MediaInfoDocumentImpl(
+			@JsonProperty("id") String jsonId,
+			@JsonProperty("labels") @JsonDeserialize(contentAs=TermImpl.class) Map<String, MonolingualTextValue> labels,
+			@JsonProperty("statements") Map<String, List<StatementImpl.PreStatement>> claims,
```
wetneb: Magnus has complained about this `statements` parameter not being consistent with the `claims` parameter in other entity types… it could potentially be changed to `claims` in the future, maybe? I wonder if it would make sense to support both in this implementation.

Tpt: Indeed, it make sense if the JSON output is changed or if people try to deserialize some custom-made JSON and uses the "claims" field without taking care.

Issue #428, opened 2019-08-10
Mising token on logout


skodapetr:
Exception:
```
Caused by: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [notoken] The "token" parameter must be set.
	at org.wikidata.wdtk.wikibaseapi.apierrors.MediaWikiApiErrorHandler.throwMediaWikiApiErrorException(MediaWikiApiErrorHandler.java:55)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.checkErrors(ApiConnection.java:466)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.sendJsonRequest(ApiConnection.java:406)
	at org.wikidata.wdtk.wikibaseapi.ApiConnection.logout(ApiConnection.java:339)
```
Code to reproduce:
```
        ApiConnection connection = new ApiConnection("https://www.wikidata.org/w/api.php");
        connection.login("username", "pswd");
        connection.logout();
```
The logout method sends following parameters:
```
assert -> user
format -> json
action -> logout
```
however the [documentation](https://www.wikidata.org/w/api.php?action=help&modules=logout) suggest that the token parameter is required as indicated by the exception.

ApiConnection is deprecated, it is suggested to use PasswordApiConnection instead. However, I was not able to find that class in this repository. I was able to find BasicApiConnection, but it suffers from the same issue. Is there any other class that should be used?

The same code works fine on our demo instance with version MediaWiki 1.32.0 but the wikidata.org is of version 1.34.0-wmf.17 (8268fd7) with a lot more extensions. 
Our version also does not require (according to documentation) the token parameter for logout.

wetneb:
Hi @skodapetr, thanks for this report. Sorry about the confusion between PasswordApiConnection and BasicApiConnection, indeed it is BasicApiConnection that we meant to recommend. This deprecation is just a first step towards a migration to a setup where ApiConnection is an interface that is implemented by various login methods (basic password, OAuth, 2FA…).

It looks like it would be good to add a token indeed (if it also works with instances where the token is not required?)
Any PR in this direction would be most welcome, if you want to give it a go.

skodapetr:
@wetneb It seems that using the token works with both instances I tested with (1.32.0, 1.34.0-wmf.17), so the change should be backward compatible.

[...]

PR #429, opened 2019-08-15
Issue 428

skodapetr:
This pull request was created to address issue #428.

The proposed solution is to move tokens management into ApiConnection class. The idea is that the ApiConnection keeps track of token requests and store them. This allows reusing the token when necessary, for example for logout operation. To reflect this fact in the API of ApiConnection  a new method *getOrFetchToken* was introduced. I'm not sure about the name, as we can keep *fetchToken* and hide the fact that it may not always cause fetch operation. I prefer a more explicit solution although they are more verbose, that is personal preference.

As tokens are managed, and cached, in ApiConnection the WbEditingAction class does not store the *CSRF* token and instead, it gets it from the ApiConnection.

The tests were also changed to reflect a fact, that *CSRF* can be obtained only after login. This seems to be original intention, based on the resource test file names but was not implemented in tests.

It seems that using the *CSRF* token works with both instances I tested with (1.32.0, 1.34.0-wmf.17), so the change should be backward compatible. But the older version API complain that the token is unknown.
```
API warning [main]: Unrecognized parameter: token.
```

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -190,6 +190,11 @@ public class ApiConnection {
 	final Map<String, String> cookies;
 
 	/**
+	 * Map of requested tokens.
+	 */
+	final Map<String, String> tokens;
+
+	/**
 	 * Mapper object used for deserializing JSON data.
 	 */
 	final ObjectMapper mapper = new ObjectMapper();
@@ -206,6 +211,7 @@ public class ApiConnection {
 	public ApiConnection(String apiBaseUrl) {
 		this.apiBaseUrl = apiBaseUrl;
 		this.cookies = new HashMap<>();
+		this.tokens = new HashMap<>();
 	}
 	
 	/**
@@ -217,6 +223,8 @@ public class ApiConnection {
 	 * 		map of cookies used for this session
 	 * @param loggedIn
 	 * 		true if login succeeded.
+	 * @param tokens
+	 * 		map of tokens used for this session
 	 */
 	@JsonCreator
 	@Deprecated
@@ -224,11 +232,13 @@ public class ApiConnection {
 			@JsonProperty("baseUrl") String apiBaseUrl,
 			@JsonProperty("cookies") Map<String, String> cookies,
 			@JsonProperty("username") String username,
-			@JsonProperty("loggedIn") boolean loggedIn) {
+			@JsonProperty("loggedIn") boolean loggedIn,
+			@JsonProperty("tokens") Map<String, String> tokens) {
 		this.apiBaseUrl = apiBaseUrl;
 		this.username = username;
 		this.cookies = cookies;
 		this.loggedIn = loggedIn;
+		this.tokens = tokens;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/BasicApiConnection.java
```diff
@@ -57,14 +57,17 @@ public BasicApiConnection(String apiBaseUrl) {
...
 	 */
 	@JsonCreator
 	private BasicApiConnection(
 			@JsonProperty("baseUrl") String apiBaseUrl,
 			@JsonProperty("cookies") Map<String, String> cookies,
 			@JsonProperty("username") String username,
-			@JsonProperty("loggedIn") boolean loggedIn) {
-		super(apiBaseUrl, cookies, username, loggedIn);
+			@JsonProperty("loggedIn") boolean loggedIn,
+			@JsonProperty("cookies") Map<String, String> tokens) {
```
wetneb: There seems to be a mismatch between the serialization key `"cookies"` and the variable name `tokens`.

skodapetr: Thanks for pointing that out, should be fixed by now.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
```diff
@@ -788,40 +782,12 @@ private EntityDocument parseJsonResponse(JsonNode entityNode) throws IOException
...
-	 * Obtains and sets a new CSRF token, whether or not there is already a
-	 * token set right now.
-	 */
-	private void refreshCsrfToken() {
-
-		this.csrfToken = fetchCsrfToken();
+		return connection.getOrFetchToken("csrf");
 		// TODO if this is null, we could try to recover here:
 		// (1) Check if we are still logged in; maybe log in again
 		// (2) If there is another error, maybe just run the operation again
```
Tpt: This comment does not look relevant anymore. I believe it should be moved to the `ApiConnection.getOrFetchToken` method. The `getCsrfToken` might also be removed now that it is just composed of one method call.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -335,6 +345,7 @@ public void logout() throws IOException {
 			Map<String, String> params = new HashMap<>();
 			params.put("action", "logout");
 			params.put("format", "json"); // reduce the output
+			params.put("token", fetchToken("csrf"));
```
Tpt: Maybe`getOrFetchToken` if the csrf token has been already loaded?

skodapetr: Yes, I probably missed that as I originally tested with this code.
One more thing to consider might the re-fresh of CSRF certificate as in ```WBEditAction```, as I'm not sure whether the close may fail with "older"? CSRF certificate.

skodapetr:
@Tpt Thanks for the comments.
I updated the pull request to reflect the comments.

I understand that every programmer has their own preferences, so please feel free to update anything to make it match the code style and chosen architecture of this repository.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java
skodapetr: Should be resolved.

Tpt:
Thank you very much!

PR #432, opened 2019-08-28
Return null in getBestStatements if there are none

bennofs:
StatementGroup requires at least one statement, so calling
getBestStatements if there are no best statements previously would throw
a "non-empty list of statements must be provided" validation error.

--- 

Another solution may be to remove the assertation that statement groups cannot be empty. 
Then we could return an empty statement group instead of null.

Having no best statement can happen if there is only a deprecrated
statement in the statement group.


wetneb:
Looks sensible to me! I would just mention that the method can return null in its documentation (especially in the interface `StatementGroup`).

bennofs:
@wetneb thanks, changed the documentation to mention that

PR #434, opened 2019-08-29
Add method to fetch property types using SPARQL

bennofs:
This method allows fetching types of properties in a single batch with a simple
SPARQL query. In contrast to the current method using the Wikidata API, this
approach only needs a single request.

--- 

This PR right now just includes the implementation of the method and does not use it by default yet. Some questions:

1. Should we use this per default for the Wikidata Property Register? If yes, when should the initial fetch happen? (Lazily on first property access?)
2. Should this be part of the PropertyRegister (as it is right now) or a separate class that extends the PropertyRegister?
3. Do we intend to keep the method using the Wikibase API? (I think yes, to support instances of Wikibase that don't have a query service?)
4. Does this need any tests, and if yes how should these tests be written? It would be nice to test directly against the online WDQS but then the tests require connectivity which is a pain and probably not worth it.

### DIFF
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/PropertyRegister.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/PropertyRegister.java
@@ -23,11 +23,21 @@ import java.io.IOException;
  */
 
 import java.util.*;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.Map.Entry;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.eclipse.rdf4j.model.IRI;
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
+import org.wikidata.wdtk.datamodel.implementation.PropertyIdValueImpl;
 import org.wikidata.wdtk.datamodel.interfaces.DatatypeIdValue;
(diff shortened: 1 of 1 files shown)


Tpt:
Thank you for this change. Have you tried the performances compared to the change you did for the API based requests to see if doing a SPARQL queries improves a lot performances? I don't think it is worth to add a dependency to a SPARQL endpoint just to earn a few seconds on a conversion process that is often very long.

> Should we use this per default for the Wikidata Property Register? If yes, when should the initial fetch happen? (Lazily on first property access?)

Maybe if it's very good for performances. Probably not if it is not the case because of the extra setup required.

> Should this be part of the PropertyRegister (as it is right now) or a separate class that extends the PropertyRegister?

I believe it's fine here.

> Do we intend to keep the method using the Wikibase API? (I think yes, to support instances of Wikibase that don't have a query service?)

Yes.

> Does this need any tests, and if yes how should these tests be written? It would be nice to test directly against the online WDQS but then the tests require connectivity which is a pain and probably not worth it.

I believe it's fine to leave it without tests if SPARQL retrieval is not the default.

bennofs:
I haven't tested it, and you are probably right that it doesn't add a lot of time if you're running a full conversion process. But, it sometimes makes sense to process a shorter dump as well. For example, I frequently do that to just quickly test something. In that case, the overhead of doing a request for almost ever y property at the beginning can be noticable.

Another advantage of using the SPARQL method is that you only need network connectivity at the start of the dump which makes it more robust when for example switching networks on a laptop while a dump is running.

So I think it's best to keep the method as is without making it the default. Then people can use it if they have those special requirements but we don't depend on a sparql service for normal operation.

bennofs:
I rebased the branch on current master to fix the conflicts. I also ran a quick test, comparing the SPARQL method with the API method on a short dump (containing only 9999 entities):

- SPARQL method: 50 seconds
- API method: 223 seconds

so for this short dump, there is a significant performance difference.

Tpt:
Thank you for the tests! There is indeed a significant performance difference. Let's not change the default for now and merge this change. Thank you!

PR #438, opened 2019-08-30
Expose RDFConverter building blocks to improve extensibility

bennofs:
The current flag-based interface to control the RDF export is quite inflexible, and cannot be extended from applications making use of wikidata toolkit, since the main functions are all package-private.

The main idea behind this patch is to expose the building blocks for the full RDF export, and let the user of the library decide how to compose them. This also has the advantage that we get a clean separation between all the generation code and the code that decides what to export based on flags.

---

This patch factors code from RdfConverter into a new AbstractRdfConverter base class.
The advantage of that approach is that extending the RDF export becomes easier:
you can create your own subclass of the AbstractRdfConverter but still reuse all
the `writeXXXX` parts to perform the RDF generation for property/item documents
and their parts.

The current RdfConverter now makes use of that to provide the previous
interface, where control of the parts to export is guided by a set of simple flags.

Using AbstractRdfConverter, much more complex filtering can be implemented by
extending the class, for example exporting terms only for some entities.

--- 

This is just a first idea, I am open to any suggestions. Also this still needs some more doc comments.

### DIFF
--- a/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/AbstractRdfConverter.java
+++ b/wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/AbstractRdfConverter.java
@@ -1,0 +1,420 @@
+package org.wikidata.wdtk.rdf;
+
+/*
+ * #%L
+ * Wikidata Toolkit RDF
+ * %%
+ * Copyright (C) 2014 Wikidata Toolkit Developers
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


Tpt:
There seems to be some errors in your change that prevents compilation (see Travis output).

Instead of relying on inheritance I would be more in favour of relying on composition by having a `StatementRdfConverter`, a `ItemRdfConverter`... with a specific instance of `StatementRdfConverter` given to the `ItemRdfConverter`. It seems to me that it would be more flexible and future proof than just relying on method overrides.

bennofs:
> There seems to be some errors in your change that prevents compilation (see Travis output).

oops, fixed now

> Instead of relying on inheritance I would be more in favour of relying on composition

I thought about ways to rely on composition, but the problem is that it isn't clear where to draw the line. On first thought, a separate ItemRdfConverter and PropertyRdfConverter makes sense, but then these two share quite some code that is "generic" for TermedStatementDocuments. But if we want to support lexemes some time in the tuture then we get another kind of top-level document that does not have terms, but is still a StatementDocument. I guess we could have a TermedDocumentConverter and a StatementDocumentConverter and so on but this would probably create so many parts that just composing them is again quite some code. 

A separate `StatementConverter` makes sense though, as we have quite some options when converting statements (how to deal with references, qualifiers etc and full vs simple). But I think that is somewhat orthogonal from this change here, so I opted to not do that change at the same time.

Tpt:
Ok! Let's merge this

PR #445, opened 2019-09-27
RDF: improve decimal and coordinate value export

bennofs:
Coordinates in the RDF produced by Wikidata for CoordinateValues are in the wrong order, according to the spec. The Wikidata RDF dump has them in the correct order. Fixing this is of course backwards incompatible, but there seems to be no other way. Or do we want to differ from the spec forever?

For decimals, we previously generated scientific notation in some cases, which is not permissible for `xsd:decimal`.


Tpt:
Indeed. Thanks!

PR #449, opened 2019-10-15
Rdf schema updates

Tpt:
Improved version of PR #444 bhy @bennofs 

This PR includes a few small changes that make the Wikidata RDF export more similar to what's in the dumps.

Some of these changes are backwards-incompatible, such as the change to how SomeValue claims are exported or the IRI changes. Not sure how we want to deal with that.

The statement id / reference hash changes are not strictly necessary, but also simple and it's nice if the subject IRIs match the full dumps.


Tpt:
@bennofs what do you think about it?

PR #451, opened 2019-10-21
Code cleanup

Tpt:
* Remove some dead code
* Makes nested classes static where possible
* Removes unused imports
* Uses specialized assert when available
* Removes explicit generics types where they could be inferred


wetneb:
Nice! Did you do this with an automated tool?

Tpt:
> Nice! Did you do this with an automated tool?

I did it in a semi automated way using IntelliJ suggestions.

PR #470, opened 2019-12-11
Add method to fetch the current lag from the Wikibase instance

### LINKED ISSUE #469: Continuous throttling policy for wdtk-wikibaseapi
Background: https://phabricator.wikimedia.org/T240442

Bots currently edit at full speed until the lag reaches 5 seconds. They should instead slow down as the lag rises. We need to introduce a continuous throttling mechanism in WTDK.

### LINKED ISSUE #510: Access current maxlag value of the server
A query with -1 as the maxLag value returns the actual server lag without any metadata.
I would find it convenient if the API could provide a simple "ping" call that would return the current server maxLag, so that we are informed how busy the server is. I could not find a way to know this besides receiving an Exception when the call fails due to maximum number of retries reached.

wetneb:
Initially designed for #469, closes #510.


wetneb:
Since there is no established guidance from WMDE about how bots should be throttled, I am removing the throttling logic to keep only the change which makes it possible to retrieve the current maxlag.

PR #483, opened 2020-02-18
Make close of asyncOutputStream synchronous in example

bennofs:
Usually, calling `close` on an output stream ensures that all data has been flushed to disk after `close` returns. This was not the case for the asyncOutputStream as the worker thread was independent. This commit fixes that issue.

--- 

This may be a somewhat special case, but as this issue is a little bit delicate and might be
not obvious to people copying the code from the example, I think it makes sense to implement
it more safely.

---

Sorry for the reformating of unchanged code, tell me if you'd like me to remove those changes.


wetneb:
Looks sensible to me, thanks!

PR #490, opened 2020-04-08
wdtk-wikibaseapi failes to login due to lower-case set-cookie session id sent from Wikidata

karlwettin:
Suddenly Wikidata is sending me header "Set-Cookie" as "Set-Cookie" and "set-cookie" mixed over multiple headers, where the previous old version of the code only handled the "Set-Cookie" header and thus skipping session id which is causing my client to fail to login. AFAIK headers are case insensitive according to the RFC, so this is not an error by Wikidata even though one could question this sudden change.

Sorry for the removed trailing end line spaces, IntelliJ did that automatically for me and I didn't notice it. The important part is at line 701
```
 -		List<String> headerCookies = headerFields
 -				.get(ApiConnection.HEADER_FIELD_SET_COOKIE);
 -		if (headerCookies != null) { 
 +		List<String> headerCookies = new ArrayList<>();
 +		for (Map.Entry<String, List<String>> headers : headerFields.entrySet()) {
 +			if (HEADER_FIELD_SET_COOKIE.equalsIgnoreCase(headers.getKey())) {
 +				headerCookies.addAll(headers.getValue());
 +			}
 +		}
 +		if (!headerCookies.isEmpty()) {
```

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -9,9 +9,9 @@ package org.wikidata.wdtk.wikibaseapi;
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


wetneb:
Thank you very much for the fix and the thorough investigation that must have lead to it! I can confirm this solves the issue for me.

To merge this we should ideally add a test demonstrating the mixed case in cookie keys.

The long term solution would be to use a higher-level HTTP library which would handle this sort of detail for us. This would solve other issues in the same go, such as #151. We have had similar problems in the past: #271.

wetneb:
@Tpt could we please merge and release this?

wetneb:
Thanks a lot @Tpt, really appreciated!

PR #496, opened 2020-05-25
OAuth support

afkbrb:
Add OAuthApiConnection , see #411 .


afkbrb:
I am thinking about removing the login method and also adding/changing the following methods

```java
// Used for anonymous requests which don't require the user to login.
// An OkHttpClient client without signing functionality is directly instantiated here.
// set loggedIn = false
public OAuthApiConnection(String apiBaseUrl)

// This one is used for OAuth and can be used for deserializing.
// An OkHttpClient client with signing functionality is directly instantiated here.
// set loggedIn = true
public OAuthApiConnection(String apiBaseUrl, consumerKey, consumerSecret, accessToken, accessSecret)

// If the loggedIn variable is set to false (the first constructor is used), reutrn ""
// Otherwise try to fetch the username via the client instance
// (We can cache the result, so we only need to fetch the name when it's null or ""). 
public String getCurrentUser()

// This one is inherited from ApiConnection.
// We override it here, make the same request as the one in getCurrentUser(), 
// if we fetch the username successfully, we know the credentials are still valid.
public void checkCredentials()

```

@wetneb Do these changes make sense to you?

wetneb:
Yes! The more we can conform to the interface, the better, I would say.

afkbrb:
@wetneb We don't have a getter for `apiBaseUrl` both in `ApiConnection` and `BasicApiConnection` even in the lastest version, but we do have the json creator [here](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java#L251) which requires the `apiBaseUrl` param. 

The serializing result [here](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-wikibaseapi/src/test/java/org/wikidata/wdtk/wikibaseapi/BasicApiConnectionTest.java#L61) in the test doesn't contain the `baseUrl` field due to the missing of the getter of `apiBaseUrl`. 

So that's a bug? I will add the getter since we'll need it for deserializing the `OAuthApiConnection`.

wetneb:
Thanks for catching that!

afkbrb:
Need more tests on real MediaWiki OAuth server.

 I modified [FetchOnlineDataExample](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-examples/src/main/java/org/wikidata/wdtk/examples/FetchOnlineDataExample.java) and [EditOnlineDataExample](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-examples/src/main/java/org/wikidata/wdtk/examples/EditOnlineDataExample.java) locally to test OAuthApiConnnection, no issues found so far.

afkbrb:
I tested with my [owner-only consumer](https://www.mediawiki.org/wiki/OAuth/Owner-only_consumers) and also with OpenRefine wikidata extension today, both worked as expected.
oauth-success

@wetneb Could you test it yourself and get it merged to [issue-268](https://github.com/Wikidata/Wikidata-Toolkit/tree/issue-268) if the code seems fine with you. It will still take some time to get issue-268 merged with the master, I'm hoping I could use `OAuthApiConnection` from the released version instead of from my local version in OpenRefine wikidata extension soon.

wetneb:
That works great for me, thanks.

PR #497, opened 2020-05-25
Creation of QuantityValues: use ItemIdValue instead of String for unit 

kanikasaini:


wdtk-datamodel/src/test/java/org/wikidata/wdtk/datamodel/helpers/DatamodelTest.java
```diff
@@ -225,6 +241,16 @@ public final void testGetQuantityValueNoBounds() {
 		assertEquals(o1, o2);
 	}
 
+	@Test
+	public final void testGetQuantityValueNoBoundsItemIdValue() {
+		BigDecimal nv = new BigDecimal(
+				"0.123456789012345678901234567890123456789");
+		ItemIdValue unit = factory.getItemIdValue("Q1", "foo");
+		QuantityValue o1 = Datamodel.makeQuantityValue(nv, unit);
+		QuantityValue o2 = factory.getQuantityValue(nv, unit.getId());
```
Tpt, line 250: Write the unit identifier using a string here (like `"http://www.wikidata.org/entity/Q1"`) might allow us to be sure that we are actually building the right identifier (and not only check that we are using `getId()`.

kanikasaini:
Thank you for the correction. I have made changes, please see if it is alright now.

PR #498, opened 2020-05-27
Merge entity guess function from PropertyRegsiter with EntityIdValueImpl

### LINKED ISSUE #424: Code duplication: determine an entity type by looking at the format of its id

snehasi:
Resolves #424

### DIFF
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java
@@ -140,7 +140,7 @@ public abstract class EntityIdValueImpl extends ValueImpl implements
 	}
 
 	/**
-	 * RReturns the entity type of the id like "item" or "property"
+	 * Returns the entity type of the id like "item" or "property"
 	 *
 	 * @param id
 	 * 		the identifier of the entity, such as "Q42"
@@ -163,7 +163,7 @@ public abstract class EntityIdValueImpl extends ValueImpl implements
 					return returnJsonEntity ? JSON_ENTITY_TYPE_LEXEME : DatatypeIdValue.DT_LEXEME;
 				}
 			case 'M':
-				return JSON_ENTITY_TYPE_MEDIA_INFO;
+				return returnJsonEntity ? JSON_ENTITY_TYPE_MEDIA_INFO : DatatypeIdValue.DT_MEDIA_INFO;
 			case 'P':
 				return returnJsonEntity ? JSON_ENTITY_TYPE_PROPERTY : DatatypeIdValue.DT_PROPERTY;
 			case 'Q':
@@ -174,7 +174,7 @@ public abstract class EntityIdValueImpl extends ValueImpl implements
 	}
 
 	/**
-	 * RReturns the entity type of the id like "item" or "property"
+	 * Returns the entity type of the id like "item" or "property"
 	 *
 	 * @param id
 	 * 		the identifier of the entity, such as "Q42"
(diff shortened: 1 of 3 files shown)


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java
```diff
@@ -143,33 +144,47 @@ public static EntityIdValue fromId(String id, String siteIri) {
...
+					return returnJsonEntity ? JSON_ENTITY_TYPE_FORM : DatatypeIdValue.DT_FORM;
 				} else if(id.contains("-S")) {
-					return JSON_ENTITY_TYPE_SENSE;
+					return returnJsonEntity ? JSON_ENTITY_TYPE_SENSE : DatatypeIdValue.DT_SENSE;
 				} else {
-					return JSON_ENTITY_TYPE_LEXEME;
+					return returnJsonEntity ? JSON_ENTITY_TYPE_LEXEME : DatatypeIdValue.DT_LEXEME;
 				}
 			case 'M':
 				return JSON_ENTITY_TYPE_MEDIA_INFO;
```
Tpt, line 166: You probably want to return here `DatatypeIdValue.DT_MEDIA_INFO`, no?

wdtk-rdf/src/main/java/org/wikidata/wdtk/rdf/PropertyRegister.java
```diff
@@ -220,21 +221,9 @@ public String getPropertyUriPattern(PropertyIdValue propertyIdValue) {
...
-			if (value.getId().contains("F")) {
-				return DatatypeIdValue.DT_FORM;
-			} else if(value.getId().contains("S")) {
-				return DatatypeIdValue.DT_SENSE;
-			}
-			return DatatypeIdValue.DT_LEXEME;
-		case 'M':
-				return DatatypeIdValue.DT_MEDIA_INFO;
-		default:
+		try{
```
Tpt, line 224: Usually we leave a space before curly braces (e.g; `try {`)

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java
```diff
@@ -143,33 +144,47 @@ public static EntityIdValue fromId(String id, String siteIri) {
...
 			case 'Q':
-				return JSON_ENTITY_TYPE_ITEM;
+				return returnJsonEntity ? JSON_ENTITY_TYPE_ITEM : DatatypeIdValue.DT_ITEM;
 			default:
 				throw new IllegalArgumentException("Entity id \"" + id + "\" is not supported.");
 		}
 	}
 
+	/**
+	 * RReturns the entity type of the id like "item" or "property"
```
Tpt, line 177: Two `R`

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/EntityIdValueImpl.java
snehasi: Yes, thank you! I've fixed it and pushed my changes.

PR #507, opened 2020-06-05
Migrate to OkHTTP to query the MediaWiki API

### LINKED ISSUE #492: Migrate to a high level HTTP library to query the MediaWiki API
We are currently using `HTTPUrlConnection` to issue HTTP requests to the MediaWiki API, at various places in the code. We should use a higher-level library which would handle variations in the responses more reliably. For instance, OkHTTP is a popular one.

The use of this low-level interface has caused serious bugs: #491, #271.

Upgrading to a high-level interface would potentially solve other issues, such as #151.

Since we are considering introducing ScribeJava as a dependency in #411, one could check if that already depends on a high-level HTTP library.

afkbrb:
Close #492 .

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -21,6 +21,7 @@ package org.wikidata.wdtk.wikibaseapi;
  */
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import okhttp3.*;
@@ -166,6 +167,7 @@ public abstract class ApiConnection {
 	/**
 	 * Getter for the apiBaseUrl.
 	 */
+	@JsonProperty("baseUrl")
 	public String getApiBaseUrl() {
 		return apiBaseUrl;
 	}
@@ -206,8 +208,16 @@ public abstract class ApiConnection {
 	 *
 	 * @return name of the logged in user
 	 */
+	@JsonProperty("username")
 	public String getCurrentUser() {
 		return username;
+	}
+
+	/**
+	 * Returns the map of tokens (such as csrf token and login token) currently used in this connection.
+	 */
+	public Map<String, String> getTokens() {
+		return Collections.unmodifiableMap(tokens);
 	}
 
 	/**
@@ -219,7 +229,7 @@ public abstract class ApiConnection {
 	public void setConnectTimeout(int timeout) {
 		connectTimeout = timeout;
 		if (timeout >= 0) {
-			updateTimeoutSettings();
+			buildClient();
 		}
 	}
 
@@ -232,11 +242,11 @@ public abstract class ApiConnection {
 	public void setReadTimeout(int timeout) {
 		readTimeout = timeout;
 		if (timeout >= 0) {
-			updateTimeoutSettings();
+			buildClient();
 		}
 	}
 
-	private void updateTimeoutSettings() {
+	private void buildClient() {
 		OkHttpClient.Builder builder = getClientBuilder();
 		if (connectTimeout >= 0) {
 			builder.connectTimeout(connectTimeout, TimeUnit.MILLISECONDS);
@@ -386,7 +396,7 @@ public abstract class ApiConnection {
 		}
 
 		if (client == null) {
-			client = getClientBuilder().build();
+			buildClient();
 		}
 		Response response = client.newCall(request).execute();
 		return Objects.requireNonNull(response.body()).byteStream();
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -161,7 +161,7 @@ public abstract class ApiConnection {
 	 * 		        .cookieJar(...);
 	 * </pre>
 	 */
-	protected abstract OkHttpClient.Builder getBuilder();
+	protected abstract OkHttpClient.Builder getClientBuilder();
 
 	/**
 	 * Getter for the apiBaseUrl.
@@ -219,7 +219,7 @@ public abstract class ApiConnection {
 	public void setReadTimeout(int timeout) {
 		readTimeout = timeout;
 		if (timeout >= 0) {
-			OkHttpClient.Builder builder = getBuilder();
+			OkHttpClient.Builder builder = getClientBuilder();
 			builder.readTimeout(timeout, TimeUnit.MILLISECONDS);
 			client = builder.build();
 		}
@@ -234,7 +234,7 @@ public abstract class ApiConnection {
 	public void setConnectTimeout(int timeout) {
 		connectTimeout = timeout;
 		if (timeout >= 0) {
-			OkHttpClient.Builder builder = getBuilder();
+			OkHttpClient.Builder builder = getClientBuilder();
 			builder.connectTimeout(timeout, TimeUnit.MILLISECONDS);
 			client = builder.build();
 		}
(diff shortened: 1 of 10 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -200,22 +209,74 @@ public void checkCredentials() throws IOException, MediaWikiApiErrorException {
...
+	 * For negative values, no timeout is set.
+	 *
+	 * @see HttpURLConnection#setConnectTimeout
+	 */
+	public void setConnectTimeout(int timeout) {
+		connectTimeout = timeout;
+		if (timeout >= 0) {
+			OkHttpClient.Builder builder = getBuilder();
+			builder.connectTimeout(timeout, TimeUnit.MILLISECONDS);
+			client = builder.build();
```
Tpt, line 241: I'm afraid there is a code logic problem here. If you call `setReadTimeout` then `setConnectTimeout` the read timeout is going to be override because a new client is built with only the connect timeout set.

afkbrb: Sorry I didn't notice that.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -20,43 +20,31 @@
...
  */
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import okhttp3.*;
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
afkbrb: IntelliJ generated it for me, if I didn't use the @NotNUll annotation, it will throw some warnings. But I'll clean it up.

afkbrb:
I found that the connect timeout and read timeout are serialized but not deserialized in the JsonCreator, and tokens are deserialized but not serialized. That seems not correct. @Tpt @wetneb Do you think we should serialize and deserialize all of them?

Tpt:
Thank you for the updates!

I believe we should definitely properly serialize and deserialize timeouts and tokens indeed.
It might be nice to also add tests for the serialization/deserialization process to make sure it works properly.

afkbrb:
Yes, I'll of course add tests for them.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -86,48 +74,55 @@
...
-	
+	protected final Map<String, String> tokens;
+
 	/**
 	 * Maximum time to wait for when establishing a connection, in milliseconds.
 	 * For negative values, no timeout is set.
 	 */
-	int connectTimeout = -1;
-	
+	protected int connectTimeout = -1;
```
Tpt, line 109: We could maybe set most of the fields here private. This way we are sure that the subclasses are using the getter/setters.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -86,48 +74,55 @@
...
 	 * True after successful login.
 	 */
-	boolean loggedIn = false;
+	protected boolean loggedIn = false;
 	/**
 	 * User name used to log in.
 	 */
-	String username = "";
-	
+	protected String username = "";
```
Tpt, line 98: Maybe dumb question: is this field actually useful for the OAuth authentification? If it is the case, what about moving it to the `BasicApiConnection`?

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -199,23 +208,90 @@ public void checkCredentials() throws IOException, MediaWikiApiErrorException {
...
+
+	/**
+	 * Sets the maximum time to wait for when establishing a connection, in milliseconds.
+	 * For negative values, no timeout is set.
+	 *
+	 * @see HttpURLConnection#setConnectTimeout
+	 */
+	public void setConnectTimeout(int timeout) {
+		connectTimeout = timeout;
+		if (timeout >= 0) {
```
Tpt, line 231: I'm a bit scared by this `if`. What about the following (quite bad...) use case:
```
client.setConnectTimeout(100);
client.setConnectTimeout(0);
```

I guess that, with the current code, the timeout would stay at `100`.

Same question for the read timeout.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
```diff
@@ -199,23 +208,90 @@ public void checkCredentials() throws IOException, MediaWikiApiErrorException {
...
+			buildClient();
```
Tpt, line 232: Maybe dump idea: what about setting `client = null` here? This way, the client would not be initalized multiple times if the API user set both the read and connect timeout.

Some question for the read timeout.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
afkbrb: We can't use private here, because this field is needed for the json creator.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
afkbrb: It's not useful for the authentification, but it's useful for getting the username via the OAuth credentials. You can see the overridden getCurrentUser() method actually uses it.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
afkbrb: The following commit should fix this.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/BasicApiConnection.java
```diff
@@ -449,4 +353,38 @@ public void logout() throws IOException, MediaWikiApiErrorException {
 			this.password = "";
 		}
 	}
+
+	private class ApiConnectionCookieJar implements CookieJar {
```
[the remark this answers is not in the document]
afkbrb: OkHTTP itself only provides an implementation of CookieJar which simply doesn't save or load cookie. After some searching, I still can't find a proper library for us here, so I'm afraid that we have to implement it ourselves. The current implementation may need to be improved, I'll spend some time to figure out how to do it properly.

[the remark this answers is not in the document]
Tpt: There is `okhttp-urlconnection` that provides a `CookieJar` based one the one from `java.net`. It's maybe the simplest solution.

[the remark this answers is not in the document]
afkbrb: Thanks to both your suggestions, I use JavaNetCookieJar and CookieManager here, without handling the cookies ourselves, the code now should look better.

afkbrb:
Oops, the tests failed on the CI server because the mock web server's host is "localhost" on the server while is "kubernetes.docker.interna" on my local machine.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
Tpt: For that you could use the `@JsonGetter` annotation on the getter methods.

PR #525, opened 2021-02-01
Attempt to install Codecov instead of Coveralls

### LINKED ISSUE #524: Fix coverage submission to Coveralls
Currently failing when PRs are submitted from a third-party fork, for security reasons.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests

wetneb:
Apparently Codecov supports GitHub Actions without the need for any secret, so migrating to it could be a simple way to fix #524.


Tpt (approved):
Great idea!

Issue #528, opened 2021-02-03
Deprecated methods


robertvazan:
There are about 20 deprecated methods. Is it time to remove them or should they be kept longer? For how long are deprecated APIs maintained?

Tpt:
There is no depreciation policy at the moment. WikidataToolkit has not moved much in the past year so I believe all deprecated methods could now be removed.

PR #531, opened 2021-02-04
Removed deprecated methods

robertvazan:
Per #528.


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/DatamodelConverter.java
```diff
@@ -685,16 +502,9 @@ public Value visit(UnsupportedValue value) {
...
-			return Collections.emptyList();
-		}
-
 		List<StatementGroup> result = new ArrayList<>(statementGroups.size());
-		for (StatementGroup statementGroup : statementGroups) {
-			if (filter.includePropertyId(statementGroup.getProperty())) {
-				result.add(copy(statementGroup));
-			}
-		}
+		for (StatementGroup statementGroup : statementGroups)
```
wetneb, line 506: So far I think we try to keep braces around all blocks, as a good practice.

wetneb (approved):

Issue #530, opened 2021-02-03
Swallowed exceptions


robertvazan:
I notice that swallowing exceptions (catching without rethrowing or wrapping) is a widespread practice in WDTK. This can sometimes result in invalid values being propagated to application or even to API calls. Can I assume that all cases of exception swallowing are bugs and fix them?

wetneb:
I think #526 is a clear improvement and I am sure there are tons of other places where we could do the same. Hard to say if it is true "everywhere", of course there is a judgment call to make…

Issue #535, opened 2021-02-12
Some basic CONTRIBUTING doc


robertvazan:
A lot of useful information can be published in project's CONTRIBUTING.md.

Right now I am wrestling with formatter settings. My usual Java style is wildly different. I have my own custom formatter configured in Eclipse. I have configured WDTK projects to use "Eclipse [built-in]" formatter instead of my own and disabled import wildcards, but I still get large diffs when organizing imports (Ctrl+Shift+O) or formatting code (Ctrl+Shift+F), even for small selections of code. For example, import wildcards seem to be used over some threshold, but I don't know what is that threshold. Publishing standard IDE settings (or referencing existing ones on the Internet) would help reduce formatting noise without having to abandon convenient IDE features.

Issue #403, opened 2018-09-25
[discussion] new architecture for Wikibase editing


wetneb:
Wikidata-Toolkit's editing features are nice but fairly limited at the moment. Arbitrary changes can be made to items and properties, but few API actions are supported (which means that the edit summaries are not as meaningful as they should be). And now that lexemes are around, we should also allow editing for these.

My PR #323 was one step in this direction but I don't think it will scale very well. I think it is time to step back and redesign the architecture of this editing module.

As a user, at the moment, the go-to entry point for editing is `WikibaseDataEditor`, and more specifically methods like this:
https://github.com/Wikidata/Wikidata-Toolkit/blob/8cd4fbb5e3a9014d06e84bd8ea2e2817e8146aea/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java#L437-L444

Beyond the lengthy arguments list, this isn't exactly nice to work with. This abstracts away in one single method a lot of different things: all the **merging** logic (how do you match the added statements to the existing ones on the item?) and the **uploading** logic (which API actions do you use to actually upload the changes?). Both of these things should be configurable:
* the merging logic depends a lot on the job at hand - very often you want to ignore qualifiers when matching statements, for instance. Or you might want to do some fuzzy matching on geographical coordinates or quantities. All these things should be configurable.
* the uploading logic is less of a problem, but it should also be configurable. Sometimes you have huge uploads so you really want to use as least API actions as possible, but in other contexts you might want to use a different strategy to get more meaningful summaries (or to make it easier for other users to revert parts of the edits, for instance).

There are various objects involved in this process:
* The state of a document: `EntityDocument`
* A set of _planned changes_ on an entity. That would store the equivalent of the arguments of the `updateTermsStatement` method, that is: a high-level description of the changes to be made, in terms of added / deleted high-level things (statements, terms, forms, …). Importantly, such classes make no reference to the current document for that entity - the data they contain has not been compared with the current document state yet.
* A family of configurable _merging strategies_: they take an `EntityDocument`, some planned changes on them, and produce the new state of the entity (an `EntityDocument` again).
* A representation of _diffs between entity states_, describing what actually changes on the document once we have applied the changes. (This representation could also be used outside an editing context, to compare revisions made by others.)
* Various _editing strategies_, which upload a diff by breaking it down into a series of API actions (possibly just one).
[...]

PR #539, opened 2021-02-23
Expose exceptions

robertvazan:
Per #530, but only for three important classes: `JsonSerializer`, `WbGetEntitiesAction`, and `WbSearchEntitiesAction`. There are some spurious exception handlers in dumpfiles/rdf that I have not touched in this pull request.

Calling code within WDTK itself already assumes exceptions. There were no null checks anywhere, so this actually fixes a bug.

This is a breaking change but a very visible one, so there would be no silent failures in apps.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbSearchEntitiesAction.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbSearchEntitiesAction.java
@@ -1,5 +1,3 @@
-package org.wikidata.wdtk.wikibaseapi;
-
 /*
  * #%L
  * Wikidata Toolkit Wikibase API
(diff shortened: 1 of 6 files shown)


wetneb (commented):
Looks like a sensible move, thanks! But the diff seems bigger than it needs to be, at a first glance.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbSearchEntitiesAction.java
```diff
@@ -42,132 +40,130 @@
...
-                results.add(ed);
-            } catch (JsonProcessingException e) {
-                LOGGER.error("Error when reading JSON for entity "
-                        + entityNode.path("id").asText("UNKNOWN") + ": "
-                        + e.toString());
-            }
-        }
-
-        return results;
-    }
```
wetneb, line 171: This seems to be a large piece of code that got reformatted in the process, it's not clear to me what changed there?

robertvazan:
I am sorry for the large diff. As I have said in #535, there are no coding standards. Most code seems to use tabs. Eclipse doesn't work so well with mixed indentation. I have therefore converted spaces in affected code to tabs. Some code has indentation changes, because try/catch was removed. GitHub fortunately allows you to [hide whitespace changes](https://github.com/Wikidata/Wikidata-Toolkit/pull/539/files?w=1).

I am touching only three classes that are relevant in lexeme I/O. This is all a preparation for #403. Of these classes, `JsonSerializer` is really essential. The other two are changed for consistency.

I should clarify that this is not really a breaking change as long as WDTK correctly parses entities returned from Wikibase API. New exceptions are thrown only if parsing fails, which would imply a bug in WDTK. All code using `Wb*Action` classes is already handling `IOException`, so the new exceptions will be caught when Wikibase data model changes. Only direct `JsonSerializer` calls require application code changes, but that is only relevant when Wikibase data model changes. If that is a serious enough problem, I can create separate set of methods in `JsonSerializer` and deprecate the old ones.

wetneb (approved):

Issue #327, opened 2018-03-05
Replace nulls by `Optional`


Tpt:
To ensure the type safety of the code it may be nice to use the Java 8 `Optional` construction where it is relevant.

Tpt:
I am not sure that we actually want to do it because Optional is not available on Android versions before the quite new 7.0.

robertvazan:
Even `Optional` wouldn't be enough.

`Optional` is an improvement as it lets API users decide what to do about missing values (fallback, null, throw). But I noticed many methods return null if they cannot choose between two possible results. In these cases, even `Optional` wouldn't provide callers with sufficient choice.

These methods just have to return `List`. To make them more convenient, it is possible to define `Ambiguous` (not sure about the name) that would inherit from `List` and also provide API inspired by `Optional` for cases when only one item is expected by the application.

```java
public interface Ambiguous<T> extends List<T> {
	static OptionalList<T> empty();
	static OptionalList<T> of(T value);
	static OptionalList<T> ofNullable(T value);
	static OptionalList<T> of(T... values);
	boolean isPresent(); // 1 or more
	boolean isUnique(); // 0 or 1
	boolean isAmbiguous(); // 2 or more
	boolean isSingle(); // exactly 1
	Optional<T> unique(); // throws if there are 2+ items
	Optional<T> uniqueOrElse(T fallback); // fallback if there are 2+ items
	Optional<T> uniqueOrElseGet(Supplier<? extends T> supplier);
	T single(); // throws if there are 0 or 2+ items
	T singleOrElse(T fallback); // fallback if there are 0 or 2+ items
	T singleOrElseGet(Supplier<? extends T> supplier);
}
```

Application interested in picking one of the values could just use `stream()` method.

So methods that query values guaranteed to be unique would return `Optional` while possibly ambiguous results would be typed `Ambiguous`.

PR #582, opened 2021-05-11
Cleanup QuantityValue interface and implementation

### LINKED ISSUE #580: Units in QuantityValues are stored internally as strings
When constructing a QuantityValue, for instance with `Datamodel.makeQuantityValue(new BigDecimal("123"), myUnit)`, the unit will internally be stored using its IRI. It will then be reconstructed using `ItemIdValueImpl.fromIri(unitIRI)` when accessed via `QuantityValue.getUnitItemId()`. If the unit is accessed multiple times, the IRI will be parsed every time into an ItemIdValue, which is perhaps a bit wasteful.

Instead, the unit object should be preserved and stored directly as such.

wetneb:
Changes QuantityValue to use Optionals instead of nulls and stores the unit as ItemIdValue directly instead of converting back and forth with Strings.

Closes #580.


Tpt (commented):
Thank you! I have two concerns:

* We removed all usages of `java.util.Optional` to keep compatibility with Android. `Optional` has been added to Android 7. I am not sure if it a reasonable requirement.
* I believe that Wikibase more or less allows any URI to be used in quantity units. Enforcing the URI to be a valid Wikibase item URI might create problems when parsing some Wikibase (Wikidata ?) dumps. I am not sure if it is a problem in practice. Do you have insights on this?

robertvazan:
@Tpt I am using `Optional` in #570. It's also requested for all API methods in #327.

Tpt:
> @Tpt I am using Optional in #570. It's also requested for all API methods in #327.

I have not thought about it. I just had a look at Android version statistics. It seems that Android 7+ has around 90% market share now so it seems reasonable to drop support for all versions below 7. It means we might also use other Java 8 APIs like Streams.

wetneb:
> We removed all usages of java.util.Optional to keep compatibility with Android. Optional has been added to Android 7. I am not sure if it a reasonable requirement.

Arf, I had the vague intuition that we were going in the reverse direction… My bad, I shouldn't have tried to tackle two issues at once (the goal was to avoid introducing new methods which would have to be deprecated soon after when Optionals are introduced).

> I believe that Wikibase more or less allows any URI to be used in quantity units. Enforcing the URI to be a valid Wikibase item URI might create problems when parsing some Wikibase (Wikidata ?) dumps. I am not sure if it is a problem in practice. Do you have insights on this?

Oops, I should have checked that! I just followed my intuition, given by the user interface, where you can only use items from the Wikibase instance as units.

I'd be interested to see what happens when you insert a QuantityValue with an unrelated unit URI with the API. Does the edit succeed? How is it rendered afterwards, especially when trying to edit it?

Similar question for the calendar URI in TimeValues actually. And I guess the globe IRI for the geographical coordinates too, but that's not even exposed by the interface at all AFAIK.

Tpt:
>>  We removed all usages of java.util.Optional to keep compatibility with Android. Optional has been added to Android 7. I am not sure if it a reasonable requirement.
> Arf, I had the vague intuition that we were going in the reverse direction… My bad, I shouldn't have tried to tackle two issues at once (the goal was to avoid introducing new methods which would have to be deprecated soon after when Optionals are introduced).

It's fine :-). I believe we can introduce again `Optional`s, Android 7+ is now much more widespread than 2+ years ago.

> I'd be interested to see what happens when you insert a QuantityValue with an unrelated unit URI with the API. Does the edit succeed? How is it rendered afterwards, especially when trying to edit it?

I don't know. I believe there was some usages in the data of URIs like `https://www.wikidata.org/wiki/QXXX` a few years ago but it's maybe clean now. Do you want to test it?

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/QuantityValue.java
```diff
@@ -64,7 +79,14 @@
...
 	 * @throws IllegalArgumentException if the unit is not "1" (no unit) or a valid item IRI
+	 * @deprecated use {@link QuantityValue.getOptionalUnit} to avoid use of null
 	 */
+	@Deprecated
 	ItemIdValue getUnitItemId();
+	
+	/**
+	 * @return the unit of this quantity as an item id value.
+	 */
+	Optional<ItemIdValue> getOptionalUnitItemId();
```
robertvazan, line 94: Just `getUnit()` would suffice.

robertvazan (commented):
Sorry for the delay. I was a bit busy. I've added some comments. Optionals in parameters are the most important issue.

wetneb:
I tried adding a quantity value with an arbitrary URL as unit and this is what I got:
```json
{
  "errors": [
    {
      "code": "modification-failed",
      "data": {
        "messages": [
          {
            "name": "wikibase-validator-not-allowed",
            "parameters": [
              "http://foo.com/bar"
            ],
            "html": {
              "*": "Illegal value: <a rel=\"nofollow\" class=\"external free\" href=\"http://foo.com/bar\">http://foo.com/bar</a>"
            }
          }
        ]
      },
      "module": "main",
      "*": "Illegal value: http://foo.com/bar"
    }
  ],
  "servedby": "mw1357",
  "*": "See https://test.wikidata.org/w/api.php for API usage. Subscribe to the mediawiki-api-announce mailing list at &lt;https://lists.wikimedia.org/mailman/listinfo/mediawiki-api-announce&gt; for notice of API deprecations and breaking changes."
}
```

That being said, when providing a syntactically valid entity URI in the Wikibase instance itself, the existence of the corrersponding entity is not checked:

image

I could create [this statement on test.wikidata.org](https://test.wikidata.org/w/index.php?title=Q106263&oldid=541127) while http://test.wikidata.org/entity/Q1062348374837438473863 obviously does not exist.

But that would not make WDTK choke anyway (since the check I introduce is syntactical only). So I think it would mostly bring WDTK in line with Wikibase, which is probably a good thing?

If you agree, I can redo a PR without the Optional business, which does add a lot more noise and is independent.

wetneb:
Replaced by #593.

PR #593, opened 2021-06-04
Store QuantityValue units as ItemIdValue instead of Strings.

### LINKED ISSUE #580: Units in QuantityValues are stored internally as strings

wetneb:
Closes #580.
Replacement for #582.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,12 +93,6 @@
 			<version>${junitVersion}</version>
 			<scope>test</scope>
 		</dependency>
-                <dependency>
-                        <groupId>org.junit.jupiter</groupId>
-                        <artifactId>junit-jupiter-engine</artifactId>
-                        <version>${junitVersion}</version>
-                        <scope>test</scope>
-                </dependency>
 		<dependency>
 			<groupId>org.hamcrest</groupId>
 			<artifactId>hamcrest-core</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,12 @@
 			<version>${junitVersion}</version>
 			<scope>test</scope>
 		</dependency>
+                <dependency>
+                        <groupId>org.junit.jupiter</groupId>
+                        <artifactId>junit-jupiter-engine</artifactId>
+                        <version>${junitVersion}</version>
+                        <scope>test</scope>
+                </dependency>
 		<dependency>
 			<groupId>org.hamcrest</groupId>
 			<artifactId>hamcrest-core</artifactId>
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/Datamodel.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/Datamodel.java
@@ -384,7 +384,7 @@ public class Datamodel {
 	public static QuantityValue makeQuantityValue(BigDecimal numericValue,
 			BigDecimal lowerBound, BigDecimal upperBound, ItemIdValue unit) {
 		return factory.getQuantityValue(numericValue, lowerBound, upperBound,
-				unit.getIri());
+				unit);
 	}
 
 	/**
@@ -397,7 +397,7 @@ public class Datamodel {
 	 * @return a {@link QuantityValue} corresponding to the input
 	 */
 	public static QuantityValue makeQuantityValue(BigDecimal numericValue, ItemIdValue unit) {
-		return factory.getQuantityValue(numericValue, unit.getIri());
+		return factory.getQuantityValue(numericValue, unit);
 	}
 
 	/**
--- a/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/QuantityValueImpl.java
+++ b/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/QuantityValueImpl.java
@@ -50,6 +50,31 @@ public class QuantityValueImpl extends ValueImpl implements QuantityValue {
 	private final JacksonInnerQuantity value;
 	
 	/**
+     * Constructor.
+     *
+     * @param numericValue
+     *            the numeric value of this quantity
+     * @param lowerBound
+     *            the lower bound of the numeric value of this quantity or null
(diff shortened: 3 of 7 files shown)


wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/QuantityValueImpl.java
```diff
@@ -49,6 +49,31 @@
...
+     */
+    public QuantityValueImpl(
+            BigDecimal numericValue,
+            BigDecimal lowerBound,
+            BigDecimal upperBound,
+            ItemIdValue unit) {
+        super(JSON_VALUE_TYPE_QUANTITY);
+        this.value = new JacksonInnerQuantity(numericValue, lowerBound,
+                upperBound, unit);
+    }
```
Tpt, line 75: There seems to be a tab vs space problem.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/QuantityValueImpl.java
```diff
@@ -171,36 +192,48 @@ public String toString() {
...
-			Validate.notEmpty(unit, "Unit cannot be empty. Use \"1\" for unit-less quantities.");
+		    this(amount, lowerBound, upperBound, parseUnit(unit));
+		}
 
-			if(lowerBound != null || upperBound != null) {
-				Validate.notNull(lowerBound, "Lower and upper bounds should be null at the same time");
-				Validate.notNull(upperBound, "Lower and upper bounds should be null at the same time");
+        protected static ItemIdValue parseUnit(String unit) {
+		    Validate.notNull(unit, "Unit cannot be null");
+            Validate.notEmpty(unit, "Unit cannot be empty. Use \"1\" for unit-less quantities.");
```
Tpt, line 208: I would feel safer to consider both null and `""` (empty string) as `"1"` and canonicalize them as such in case there is slightly wrong JSON out there. It seems fairly unambiguous to me. But it's maybe not a good idea to depart from Wikibase if it rejects these values.

wetneb: I agree, I initially did that and then realized the tests were enforcing the current behaviour so I thought I should better keep it… It does feel like a safe move though.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/implementation/QuantityValueImpl.java
Tpt: Thank you for the fix! However, it seems you included a lot of unrelated files in this change.

pom.xml
```diff
@@ -93,6 +93,12 @@
 			<version>${junitVersion}</version>
 			<scope>test</scope>
 		</dependency>
+                <dependency>
+                        <groupId>org.junit.jupiter</groupId>
+                        <artifactId>junit-jupiter-engine</artifactId>
+                        <version>${junitVersion}</version>
+                        <scope>test</scope>
+                </dependency>
```
Tpt, line 101: It does not seems to be related to this PR topic.

Tpt (commented):
Thank you!

pom.xml
wetneb: The pom.xml change yes, but I think the other changes are actually related and pretty much required, no? Otherwise by creating a QuantityIdValue from the Datamodel helper, you will first convert an ItemIdValue to a String and then back to an ItemIdValue.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/Datamodel.java
```diff
@@ -397,7 +397,7 @@ public static QuantityValue makeQuantityValue(BigDecimal numericValue,
 	 * @return a {@link QuantityValue} corresponding to the input
 	 */
 	public static QuantityValue makeQuantityValue(BigDecimal numericValue, ItemIdValue unit) {
-		return factory.getQuantityValue(numericValue, unit.getIri());
+		return factory.getQuantityValue(numericValue, unit);
```
wetneb, line 400: To avoid this conversion to String one needs to update the DataObjectFactory interface too, hence the other changes.

pom.xml
Tpt: Yes, indeed. I was mentioning this pom.xml change.

Tpt (approved):

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/DataObjectFactory.java
```diff
@@ -254,9 +268,27 @@ QuantityValue getQuantityValue(BigDecimal numericValue,
 	 *            the unit of this quantity, or the empty string if there is no
 	 *            unit
 	 * @return a {@link QuantityValue} corresponding to the input
+	 * @deprecated use {@link #getQuantityValue(BigDecimal, BigDecimal, BigDecimal, ItemIdValue)}
```
robertvazan, line 271: @wetneb I've got some warnings after merging this. Could you please look at it? Firstly, `@deprecated` in javadoc and `@Deprecated` annotation should be used together. Secondly, existing calls (e.g. the one in `DatamodelConverter` must use the new API as otherwise they will trigger warnings.

wetneb: Thanks for noticing, see #594

PR #670, opened 2022-04-01
Add support for file uploads to ApiConnection.

### LINKED ISSUE #638: Support for file uploads
WDTK supports creating new items, properties and lexemes, but not MediaInfo entities, and for a good reason: those cannot be created without uploading the corresponding file (to Commons at least).

Would it be appropriate to add support for file upload, via [the corresponding MediaWiki API action](https://commons.wikimedia.org/w/api.php?action=help&modules=upload)? That API action has nothing to do with Wikibase itself, so I could understand that it is considered out of scope, but at the same time it does feel like we should be able to reuse an `ApiConnection` from WDTK to also perform this action.

More generally, we could potentially think about how to make it easier to reuse an `ApiConnection` object for MediaWiki API actions which are not directly supported by WDTK. Intuitively, it can make sense to use most of the MediaWiki API actions within the context of Wikibase, and there is no chance we implement a wrapper for every single of those actions in WDTK any time soon, right? Perhaps the right way to phrase this would be to look for MediaWiki Java clients and see if there could be some sort of interoperability.

wetneb:
Closes #638.

This makes the ApiConnection classes reusable to perform API calls which require uploading a file. Adding an actual wrapper on top of those actions is out of scope for WDTK, but this makes it possible for users to implement that on their side, without having to hack the user credentials out of the ApiConnection object.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -20,6 +20,7 @@
 
 package org.wikidata.wdtk.wikibaseapi;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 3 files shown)


Tpt (approved):
I have not tried to run the code but it looks ok and well tested.

PR #705, opened 2022-07-05
Add explicit dependency to okhttp in wdtk-wikibaseapi module

wetneb:
For #600


wetneb:
Thanks! I have never used <dependencyManagement> so I am not so clear on the benefits it provides, but let's talk about this further in https://github.com/Wikidata/Wikidata-Toolkit/issues/706

PR #707, opened 2022-07-06
Add clientLogin method to BasicApiConnection

Lhaaits:


wetneb (commented):
Hey this looks really great, nice refactoring with your big `enum` (did not know one could do that!)

I haven't tested it yet - did you test it on your side, and with which Wikibase instance? I remember looking into this earlier and it felt quite complicated because this login workflow can involve 2FA verifications. It does not look like those are supported yet, are they? Do you think this would be doable? Not sure what the Java API would like for our users.

Lhaaits:
By the way, do you know where all possible login error values can be found? I wasn't able to reproduce all possible values to complete the enum. For example LOGIN_THROTTLED or LOGIN_BLOCKED. I tried looking for a repo for the api.php, but no luck yet. And the docs are also not that extensive.

Lhaaits:
@guenthermi I see you added the login error values originally. Do you by any change remember what you based those values on?

wetneb:
@Lhaaits I would not block this PR for that. On my side I was waiting for this discussion about error codes to be resolved, but if that is stalled I would be very fine with merging this as is.

robertvazan:
@Lhaaits I do not contribute to server-side code. Sorry I cannot be of more help.

Lhaaits:
@robertvazan Fair enough. Thanks for the quick reply :)
@wetneb I would like to merge if that's ok. No luck finding the error codes unfortunately, except for the ones I could reproduce locally.

wetneb:
@Lhaaits do you need to be released soon? That could probably be a new minor version (`0.14.0`).

Lhaaits:
@wetneb Yeah a minor makes sense :)

PR #719, opened 2022-08-10
Deprecate DatatypeIdValue::getIri()

### LINKED ISSUE #718: Remove datatype IRI from Wikibase data model implementation
Datatype IRIs are not native to the Wikibase Data Model. They crease a lot of problems like #716.

We might replace the  `DatatypeIdValue` IRI field with a new field storing the JSON datatype ID.

The IRI support would be moved to the RDF module and stay there.

Migration path:
1. Merge #717
2. Deprecate the IRI field of the DatatypeIdValue
3. After one or two release, drop the IRI field.
4. Migration done!

wetneb:
Closes #718.


Tpt (approved):
Thank you!

PR #746, opened 2022-11-22
Fix fetching mids of multiple non-existent files.

### LINKED ISSUE #745: Incorrect behavior when converting multiple non-existent filenames to mids
When giving multiple non-existent filenames to convert to Mids, `WikibaseDataFetcher.getMediaInfoIdsByFileName` does not return the expected value because the API returns a payload of the form:
```json
{"batchcomplete":"",
  "query":{
    "pages":{
         "-1":{"ns":6,"title":"File:DoesNotExist.png","missing":""},
         "-2":{"ns":6,"title":"File:DoesNotExistEither.jpg","missing":""}
}}}
```

The code assumes that non-existent filenames always get `-1` as key, but the second gets `-2` so code returns a "mid" of `M-2`.

wetneb:
Closes #745.

I would release this immediately after merge as this is the source of an important bug in OpenRefine.


Tpt (approved):
I should have not let this slip during the first review. It's not the first time I see this problem... Thank you!

wetneb:
Oh no problem at all - I would not have caught that myself (and had no idea you reviewed this change)

PR #748, opened 2022-12-01
Fix deserialization of lexemes to accept bogus JSON produced by Wikibase

### LINKED ISSUE #568: JSON dump parsing error while running examples
I encountered below error message when running the examples `GreatestNumberProcessor` and `GenderRatioProcessor`.   
```
Problematic line was: {"type":"lexeme","id":"L34069","lemmas":{"de":{"la...
Error when reading JSON for entity: Missing type id when trying to resolve subtype of [simple type, class org.wikidata.wdtk.datamodel.implementation.FormDocumentImpl]: missing type id property 'type' (for POJO property 'forms')
 at [Source: (String)"{"type":"lexeme","id":"L34070","lemmas":{"de":{"language":"de","value":"Zahnl\u00fccke"}},"lexicalCategory":"Q1084","language":"Q188","claims":{"P5185":[{"mainsnak":{"snaktype":"value","property":"P5185","datavalue":{"value":{"entity-type":"item","numeric-id":1775415,"id":"Q1775415"},"type":"wikibase-entityid"},"datatype":"wikibase-item"},"type":"statement","id":"L34070$B9D7523B-8AE6-4C39-85A2-EE7CC71C5228","rank":"normal"}],"P8376":[{"mainsnak":{"snaktype":"value","property":"P8376","datavalue""[truncated 4073 chars]; line: 1, column: 1043] (through reference chain: org.wikidata.wdtk.datamodel.implementation.LexemeDocumentImpl["forms"]->java.util.ArrayList[0])
```
The dump is automatically downloaded `20210407.json.gz`.  
  
I tried to run the code on Windows (Build with Maven with IDEA, JDK 15.0.1) and Ubuntu (openJDK 1.8.0_282, run as IDEA remote). 

It seems the dump file size is only ~200MB. Would that be correct?

Any help is appreciated. Thank you.

wetneb:
Closes #568.


wetneb:
The right thing would of course to fix this in Wikibase instead, but given their stable policy interface for Wikidata it looks like a pretty daunting process.

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/LexemeDeserializer.java
```diff
@@ -0,0 +1,98 @@
...
+                        typeFactory.constructCollectionType(List.class, StatementImpl.PreStatement.class)));
+
+        List<FormDocument> forms = deserializationContext.readTreeAsValue(
+                object.get("forms"),
+                typeFactory.constructCollectionType(List.class, FormDocumentImpl.class));
+
+        List<SenseDocument> senses;
+        if (object.get("senses") instanceof ObjectNode && ((ObjectNode)object.get("senses")).isEmpty()) {
+            // case for https://github.com/Wikidata/Wikidata-Toolkit/issues/568
+            senses = Collections.emptyList();
```
Tpt, line 74: Genuine question: couldn't this bug also happen for "forms"?

Tpt (approved):
Thank you! PHP with its empty array = empty object is so much fun...

wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/helpers/LexemeDeserializer.java
wetneb: I thought that lexemes always have at least a form, so that list could not be empty, but I am not sure where I dreamed that up… Sigh! At least the custom deserializer will be a bit more worth the boilerplate.

PR #764, opened 2023-03-08
Add uselang parameter for wbSearchEntities

Lhaaits:


Lhaaits:
I'm a bit confused. It seems that a pull request that I created a while ago, with the same changes was already merged to master, but for some reason github still sees these changes compared to my fork, so I could open another pull request. I'll close this one

Issue #442, opened 2019-09-23
CSRF Token error


luisenriqueramos1977:
Dear friends, 

I have been working with wikidata for sometime, but since last week , when I try to make a bulk load, I receive a CSRF error. 

I mean, when I edit wikidata and write the item, I do not receive any error message and write the item, but now I receive a CSRF Token error message as shown below:

```
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
Exception in thread "main" 2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
2019-09-23 09:12:42 ERROR - Gave up after several retries. Last error was: org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
org.wikidata.wdtk.wikibaseapi.apierrors.TokenErrorException: [badtoken] Invalid CSRF token.
[...]

PR #765, opened 2023-03-09
442 reset tokens

Lhaaits:
Related to https://github.com/Wikidata/Wikidata-Toolkit/issues/442.
It seems that clearing the login token when encountering this exception solves the "bad CSRF token" message, because a new token is fetched.


Tpt (approved):
Thank you!

PR #776, opened 2023-05-05
Add release instructions

wetneb:
This is an attempt to codify the practice I have been following recently. Feel free to improve!


wetneb:
@Lhaaits let me know if anything is unclear! The goal would be that you are able to do a release on your own.

wetneb:
I have just tested the workflow and it works as expected, therefore merging. I hope this empowers others to make releases as I would like to move away from this task.

PR #796, opened 2023-06-30
Wikibase API: Return revision id after making an edit.

### LINKED ISSUE #795: Editing API: return revision id on successful edit
The main method to make edits to existing entities looks like this:
```java
public void editEntityDocument(
	EntityUpdate update, boolean clear, String summary, List<String> tags)
        throws IOException, MediaWikiApiErrorException;
```
It would be convenient if it could instead return some information about the edit (if it was made successfully). For instance, the new revision id.

wetneb:
Closes #795.

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
@@ -1,0 +1,70 @@
+package org.wikidata.wdtk.wikibaseapi;
+
+/*-
+ * #%L
+ * Wikidata Toolkit Wikibase API
+ * %%
+ * Copyright (C) 2014 - 2023 Wikidata Toolkit Developers
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import java.util.Objects;
+
+/**
+ * Holds information about a successful edit made via {@link WikibaseDataEditor}.
+ * The state of the entity after edit is not provided here because it is not possible
+ * for WDTK to determine it reliably from the response of the server. Indeed, it is
+ * possible that the data in the entity after the edit differs from the data in the
+ * entity before the edit plus the changes of the edit itself, because it can be that
+ * another edit touched independent parts of the entity. This can happen even if the base
+ * revision id is provided.
+ */
+public class EditingResult {
+    
+    private final long revisionId;
+
+    public EditingResult(long revisionId) {
+        super();
+        this.revisionId = revisionId;
+    }
+
+    /**
+     * The identifier of the revision of the last edit made by the editing action.
+     * If no edit was made, 0L is returned.
+     */
+    public long getLastRevisionId() {
+        return revisionId;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(revisionId);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null) return false;
+        if (getClass() != obj.getClass()) return false;
+        EditingResult other = (EditingResult) obj;
+        return revisionId == other.revisionId;
+    }
+
+    @Override
+    public String toString() {
+        return "EditingResult [revisionId=" + revisionId + "]";
+    }
+    
+}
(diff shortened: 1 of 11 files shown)


wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
```diff
@@ -0,0 +1,70 @@
...
+        if (getClass() != obj.getClass()) return false;
+        EditingResult other = (EditingResult) obj;
+        return revisionId == other.revisionId;
+    }
+
+    @Override
+    public String toString() {
+        return "EditingResult [revisionId=" + revisionId + "]";
+    }
+
```
robertvazan, line 69: Just a suggestion: It might be a good idea to use [`EqualsBuilder`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/EqualsBuilder.html) and [`ToStringBuilder`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ToStringBuilder.html) from Apache Commons (and perhaps their `reflectionEquals` and `reflectionToString` methods) across WDTK code if Java records are too far in the future. It minimizes bugs at the cost of a little performance overhead.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java
```diff
@@ -530,13 +533,13 @@ public ItemDocument editItemDocument(ItemDocument itemDocument,
...
 	 */
-	public void editEntityDocument(
+	public EditingResult editEntityDocument(
 			EntityUpdate update, boolean clear, String summary, List<String> tags)
 			throws IOException, MediaWikiApiErrorException {
 		long revisionId = update.getBaseRevisionId();
 		if (!clear) {
 			if (update.isEmpty())
-				return;
+				return new EditingResult(0L);
```
robertvazan, line 542: Shouldn't this be the current revision ID rather than zero?

robertvazan (approved):
Looks alright except the zero revision ID. I did't check the tests. I trust you have made sure this works with the actual Wikibase API.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java
wetneb: I considered that but did not go for it because the update might not have a base revision id in the first place. I want to be able to tell from the response if an edit was made or not.

I guess we could forward the base revision id and add another field on `EditingResult` to record the number of edits (= API actions) used to carry out the update. At the moment this would be either 0 or 1, but there could be scenarios where we cannot or do not want to do an update atomically (for instance to get more meaningful edit summaries).

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
wetneb: I did not know about those. When writing this class I could simply use `@Data` from Lombok:
https://projectlombok.org/features/Data
Overall I think there could be quite some benefits to using Lombok in WDTK (but I have not used it in any real project yet). What do you think?

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.java
robertvazan: Ah, okay, I thought it's an accidental omission. Now it's clear there's reasoning behind it. You might want to consider optionals, so that people can provide their own fallback in `orElse()`.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
robertvazan: I used Lombok in my projects for some time and then dropped it due to severe compatibility problems that were blocking upgrades. Lombok has to rely on internal compiler APIs and it therefore requires specialized integration with every compiler (and thus every IDE) and sometimes breaks when compiler is upgraded. Plus anyone working on the project has to manually add Lombok to their IDE configuration (see the Install menu on Lombok website). It's less compatible than just using Java 17 and records.

wetneb: Oh great, then that's saving me a lot of effort by not looking into it.

wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/EditingResult.java
```diff
@@ -41,11 +42,11 @@ public EditingResult(long revisionId) {
...
     /**
-     * The identifier of the revision of the last edit made by the editing action.
-     * If no edit was made, 0L is returned.
+     * The identifier of the revision of the last edit made by the editing action,
+     * if any edit was made.
      */
-    public long getLastRevisionId() {
-        return revisionId;
+    public Optional<Long> getLastRevisionId() {
+        return revisionId == 0 ? Optional.empty() : Optional.of(revisionId);
```
robertvazan, line 49: Why not use [`OptionalLong`](https://docs.oracle.com/javase/8/docs/api/java/util/OptionalLong.html)?

wetneb: Simply wasn't aware of its existence!

PR #866, opened 2024-04-09
Switch to using <dependencyManagement> for Maven dependencies

### LINKED ISSUE #706: Migrate to <dependencyManagement> to control versions of dependencies
We currently use Maven properties in the root POM to specify the versions of our dependencies, but that only works to control the versions of the dependencies that are directly mentioned in submodules. Using the <dependencyManagement> section would instead let us control the versioning of transitive dependencies as well.

Is that correct @robertvazan?

wetneb:
Closes #706.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,49 +72,139 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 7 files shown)


robertvazan (approved):
Looks alright to me.

Issue #900, opened 2024-08-09
Adopt "conventional commits" message style and release-please workflow


wetneb:
I recently discovered [release-please](https://github.com/googleapis/release-please) which eases the release process by automating quite a lot of [the tasks that we currently do manually](https://github.com/Wikidata/Wikidata-Toolkit?tab=readme-ov-file#how-to-make-a-release). I would be keen to try out this tool and see if it could work for Wikidata-Toolkit, which has a fairly low traffic.

This would require adopting the "[conventional commits](https://www.conventionalcommits.org/)" style of commit message:

> *  `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
> *  `feat:` which represents a new feature, and correlates to a SemVer minor.
> *  `feat!:`, or `fix!`:, `refactor!:`, etc., which represent a breaking change (indicated by the !) and will result in a SemVer major.

I think it should be doable!

PR #912, opened 2024-09-19
Add release-please workflow for #900

wetneb:
Sadly the PAT I generated is only valid for one year (maximum duration) so that would need renewing yearly, let's see.


wetneb:
Let's try this out, happy to revert if it doesn't work for people (but I seem to be the only active person here at the moment anyway)

PR #930, opened 2024-11-28
Refactored implementation and Design smells

abukalam2909:
The following implementation smells has been refactored:

- Complex conditional
- Complex Method
- Long Identifier

The following design refactorings have been made:

- Extract Class (Moved copy methods in DatamodelConverter to new class CopyUtils)
- Pull-up method (Moved hashCodeForTermedDocument in hash class to TermedDocument interface)
- Move field (Moved Logger initialization into constructor in DatamodelConverter class to make it more cohesive)


wetneb:
Hi @abukalam2909, I am curious what prompted you to submit those changes. Could you elaborate on what brought you to working on this?

I had a brief look at the changes and it's not clear to me that they are really useful.

PR #960, opened 2025-08-28
Fix #959: Configurable User-Agent with versioned default

### LINKED ISSUE #959: Provide a way to set user agent on requests
Wikidata is not rejecting all requests without a user agent header set, so all users of this toolkit will be blocked.

It would be very desirable to provide a facility for callers to set a custom user agent string which could be used on all requests.

rzo1:
This PR introduces support for configuring a custom User-Agent header in HTTP requests.
When no custom User-Agent is defined, the library now falls back to a meaningful default.

I verified the change with one of our applications:

On versions prior to **0.16.1-SNAPSHOT**, the issue described in #959 happens.

Using a SNAPSHOT built from this branch, the behavior works correctly for my tested use case.

### DIFF
--- a/wdtk-wikibaseapi/pom.xml
+++ b/wdtk-wikibaseapi/pom.xml
@@ -15,6 +15,10 @@
 	<name>Wikidata Toolkit Wikibase API</name>
 	<description>Java library to access the Wikibase Web API</description>
 
+    <properties>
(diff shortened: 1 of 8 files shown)


wetneb (approved):
Thank you very much for tackling this! It looks perfect to me.

PR #963, opened 2025-08-29
feat(http): Enable reporting of source data on JSON parse failures 

### LINKED ISSUE #962: Enable reporting of source data on JSON parse failures
The stack trace below was reported recently with #959

```
09:14:24.221 [                   refine] POST /command/wikidata/login (0ms)
09:14:24.337 [       connection_manager] Unrecognized token 'Please': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 8] (115ms)
org.wikidata.wdtk.wikibaseapi.LoginFailedException: Unrecognized token 'Please': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 8]
        at org.wikidata.wdtk.wikibaseapi.BasicApiConnection.login(BasicApiConnection.java:193)
        at org.wikidata.wdtk.wikibaseapi.BasicApiConnection.login(BasicApiConnection.java:155)
```

I suspect "Please" is the beginning of some plain text error message like "Please set your User-Agent header." It would be easier for users to identify cases like this if they had the full text (or the contents of the HTML error page when those get returned in the place of JSON). I don't think there's sensitive data being returned from the Wikidata APIs, so it would be nice to enable Jackson's `StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` feature.

rzo1:
This PR refactors our HTTP handling and improves error diagnostics when the server returns non-JSON responses (e.g., plain text or HTML error pages).

Previously, these cases resulted in opaque JSON parse errors without showing the actual server message or HTTP response code. With this change, users will immediately see the full context (status, headers, body) and can identify issues like missing headers, rate limiting, or incorrect parameters.

Fixes #962

### DIFF
--- a/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
+++ b/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/ApiConnection.java
@@ -36,6 +36,9 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.StreamReadFeature;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -151,7 +154,9 @@ public abstract class ApiConnection {
 	/**
 	 * Mapper object used for deserializing JSON data.
 	 */
-	private final ObjectMapper mapper = new ObjectMapper();
+    private final ObjectMapper mapper = JsonMapper.builder()
+            .enable(StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION)
+            .build();
 
 	/**
 	 * Creates an object to manage a connection to the Web API of a Wikibase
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Looks good to me. The `sendRequest2` method name isn't ideal, but I also don't have something better in mind…
Leaving open for a bit longer to leave time for others to chime in.

rzo1:
> Looks good to me. The `sendRequest2` method name isn't ideal, but I also don't have something better in mind…
> Leaving open for a bit longer to leave time for others to chime in.

Yes. I don't like the name either. We can also opt to break backward compatibility by renaming and dropping the original method (sadly, it was public).

wetneb:
Another option would be to name it something a bit scary, like `sendRequestWithoutErrorHandling`, which would serve two purposes:
* having a name different than `sendRequest`, for backwards compatibility purposes
* giving a more explicit warning that the version with error handling should likely be favored

wetneb:
Ah I see, I was referring to the difference mentioned in the doc comment in it advising to use `sendJsonRequest` instead… But you're right that the error-handling isn't really the the distinction between the two, it's rather the JSON parsing.

I also think it's okay to do this as a breaking change. It should give a compilation error that people can recover from, if they happen to use this rather internal-looking method.

rzo1:
> I also think it's okay to do this as a breaking change. It should give a compilation error that people can recover from, if they happen to use this rather internal-looking method.

I followed that path, so it should now be clear and concise.

mawiesne:
> I followed that path, so it should now be clear and concise.

I agree, full ack

PR #977, opened 2025-11-22
chore!: Update to Jackson 3 + Java 17 new baseline

don-vip:
Update to Jackson 3 + Java 17 as new baseline.

Reference:
- https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0
- https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md
- https://spring.io/blog/2025/10/07/introducing-jackson-3-support-in-spring
- https://github.com/don-vip/Wikidata-Toolkit/actions/runs/19602442390

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ 11, 17, 21 ]
+        java: [ 17, 21, 25 ]
 
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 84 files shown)


wetneb (commented):
Thanks for working on this big update! It looks good to me, but I want to leave the PR open for a while to let others chime in.

I'm curious, is WDTK being used in JOSM, or are you coming from another project?
Is your motivation primarily to avoid pulling in two copies of Jackson, or is this migration functionally required on your side (for instance because you'd be serializing with Jackson 3 a bean of your own making that contains a WDTK object, relying on its jackson serializability rather than going through `DatamodelMapper`? But I guess we do leak jackson classes in exception types (at least) anyway…)

I also wonder if we should expect to continue maintaining a branch of WDTK with Jackson 2…

Tpt:
@wetneb yes!

wetneb (approved):

PR #1006, opened 2026-05-31
governance: remove Lhaaits as owner

wetneb:
@Lhaaits did not accept the GitHub invite to the Wikidata-Toolkit org, so I propose to update this list to reflect that.

@Lhaaits if you'd like to stay let me know and I can re-issue the invite.


wetneb:
Merging this after a week has passed. Again, I would be very happy to have you onboard again, @Lhaaits.

Lhaaits:
@wetneb I don't really use it for work anymore and don't have much free time for it either. So I think it's a good move to clean up :)

